PR #395, opened 2014-03-10
dal.py: attribute "unique" added for tables

gitanoqevaporelmundoentero:
The 'unique' attribute is also added in dal.py. In this way, alternative keys different than the primary key can be included in the 'unique' attribute when defining the database tables, for example:

db.define_table('variant',
                Field('id', 'integer'),
                Field('chromosome_id', 'integer'),
                Field('genome_id', 'integer'),
                Field('start', 'integer'),
                Field('reference_allele', 'string'),
                Field('variant_allele', 'string'),
                Field('type', 'string'),
                Field('created', 'datetime'),
                Field('updated', 'datetime'),
                Field('user_id', 'integer'),
                Field('flanking_sequence', 'string'),
                primarykey=['id'],
                unique=['chromosome_id', 'genome_id', 'start', 'reference_allele', 'variant_allele'],
                migrate=False)

Changes in validate_and_update, update_or_insert, _insert and validate_and_update_or_insert make it possible for the api to insert or update records only by passing the alternative keys (those defined in the 'unique' attribute) to the update or insert method, and finally getting the primary key from the updated or inserted record.

### DIFF
--- a/gluon/dal.py
+++ b/gluon/dal.py
@@ -201,7 +201,7 @@ CALLABLETYPES = (types.LambdaType, types.FunctionType,
 TABLE_ARGS = set(
     ('migrate','primarykey','fake_migrate','format','redefine',
      'singular','plural','trigger_name','sequence_name','fields',
-     'common_filter','polymodel','table_class','on_define','rname'))
+     'common_filter','polymodel','table_class','on_define','rname', 'unique'))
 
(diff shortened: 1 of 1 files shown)


mdipierro:
will be addressed in pydal: https://github.com/web2py/pydal/issues/116

PR #432, opened 2014-04-29
commented out bare setting in AD for username allowing a user@domain.com...

murdochr:
... to import ldap groups properly.


BuhtigithuB:
I doubt that this single line commit resolve the issue of double user creation of ldap contrib... I surely had try this when I try to fix that issue (see mailing list about that)...

Richard

mdipierro:
This is a very old request. We need to decide what to do with it. This worked under the assumption that if a user logs in with pinco@pallino.com, then pinco is the username. If we change it, people will no longer be able to login in their current accounts. Anyway, let's assume we use "pinco@pallino.com" as username. Are we sure that @ is always allowed in a username for all ldap modes? have you tried this?

mdipierro:
sorry. unless the original author can explain this PR I will have to close it. I cannot test it.

PR #548, opened 2014-11-26
Fix for search after sort in SQLFORM.grid (over ajax) #2020

dexif:
https://code.google.com/p/web2py/issues/detail?id=2020


niphlod:
after 22 days I'd say either close it and live with it until the next bug-report or pack an app to reproduce the issue.

niphlod:
ok, super-closing this as there are no news. @dexif : feel free to open a thread on web2py-users with an app to reproduce the issue with current trunk

PR #580, opened 2015-01-03
Fix expression evaluation with postgres

ilvalle:


gi0baro:
Why you're not proposing this on pydal?

ilvalle:
an user needs this fix, given that it would much simpler for him to update
web2py trunk
I'll update pydal too

 Paolo

2015-01-03 14:32 GMT+01:00 Giovanni Barillari <notifications@github.com>:

gi0baro:
Ok, please add the PR also on pydal so we won't forget it ;-)

mdipierro:
I had merged this but I am having second thoughts... Will this break existing tables created with dal.py that use 0,1 for booleans on postgresql?

ilvalle:
No, this fix is for expressions evaluated at python level broken for
postgres (at least). It doesn't touch the true/false field representation.
In addition It would be necessary to check if there are other db engines
that don't accept boolean expressions as bit (0/1)

 Paolo

2015-01-07 20:28 GMT+01:00 mdipierro <notifications@github.com>:

mdipierro:
OK. Thanks.

PR #581, opened 2015-01-03
added support for POST variable passing to LOAD function

ShySec:
A new post_vars parameter allows controllers to pass POST varaibles through LOAD as appropriate. GET variables are currently passed by default (via vars).

### DIFF
--- a/gluon/compileapp.py
+++ b/gluon/compileapp.py
@@ -127,7 +127,7 @@ class mybuiltin(object):
 def LOAD(c=None, f='index', args=None, vars=None,
          extension=None, target=None, ajax=False, ajax_trap=False,
          url=None, user_signature=False, timeout=None, times=1,
-         content='loading...', **attr):
+         content='loading...', post_vars=Storage(), **attr):
     """  LOADs a component into the action's document
 
     Args:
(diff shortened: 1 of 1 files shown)


ShySec:
Integration test failed independent of this patch (server was unable to connect to the MongoDB instance)

niphlod:
what's the usecase for this ? every ajax machinery is rigged to behave using GET, and I think it should stay that way. Why do you specifically need POST ?

ShySec:
This change supports non-AJAX interfaces and allows LOADing pages with (named) forms. This makes static LOAD much more accessible and usable as a compositing tool.

For example, a dashboard or tabbed interface can be made from simply LOADing several individual views. This change permits those LOADed views to include fully interactive SQLFORM-style elements by passing the POST variables.

niphlod:
you can already with GET, and you don't need a POST to generate a form.

ShySec:
I created the following controller/test.py to demonstrate:

---
def old(): return LOAD('test','load')
def new(): return LOAD('test','load',post_vars=request.post_vars)

def load():
	form=SQLFORM.factory(Field('field'))
	if form.validate(keepvalues=True):
		return 'sent: %s'%form.vars.field
	return form
---

Loading /app/test/old shows the form, but submitted values won't validate since LOAD doesn't pass POST variables. /app/test/new fixes the problem by explicitly passing POST variables.

ilvalle:
@ShySec could you document your contribution on the book https://github.com/mdipierro/web2py-book? Create an issue or better a PR

PR #584, opened 2015-01-10
GoogleSQLAdapter: Fix NDB orderby without limitby broken in 2.9.6

anssih:
Commit 8d648f6137cea5 ("restore beahvior in gae that if no limitby,
returns iterator") made select_raw() store the query iterator in a
wrong variable ("rows", which is unused, should be "items"), causing
unsorted results to be returned.

Fix the assignment.


gi0baro:
Would be good to have this also on https://github.com/web2py/pydal

PR #587, opened 2015-01-13
fix import issue for gluon classes (ie: from dal import DAL)

ilvalle:
This PR makes

* the following two import to load the same class
```
from dal import DAL
from gluon import DAL
```

* allows users to have an app-module called as one in gluon/* without conflicts (i.e.: tools.py)

* allows us to have a contrib module which internally has an import to other internal module called as one in gluon/* (This weird case happens for simplejson which has a module called decorator)

### DIFF
--- a/gluon/custom_import.py
+++ b/gluon/custom_import.py
@@ -41,7 +41,7 @@ class CustomImportException(ImportError):
 
 def custom_importer(name, globals=None, locals=None, fromlist=None, level=-1):
     """
-    web2py's custom importer. It behaves like the standard Python importer but 
+    web2py's custom importer. It behaves like the standard Python importer but
     it tries to transform import statements as something like
     "import applications.app_name.modules.x".
     If the import fails, it falls back on naive_importer
@@ -67,6 +67,13 @@ def custom_importer(name, globals=None, locals=None, fromlist=None, level=-1):
         import_tb = None
         try:
             try:
+                if ('gluon' not in name and not name.startswith('.') and
+                   'direct_import' not in locals):
+                    try:  # Try to load a gluon class
+                        oname = 'gluon.%s' % name
+                        return NATIVE_IMPORTER(oname, globals, locals, fromlist, level)
+                    except:
+                        pass
                 oname = name if not name.startswith('.') else '.'+name
                 return NATIVE_IMPORTER(oname, globals, locals, fromlist, level)
             except ImportError:
@@ -102,6 +109,7 @@ def custom_importer(name, globals=None, locals=None, fromlist=None, level=-1):
             if import_tb:
                 import_tb = None
 
+    locals['direct_import']=True
     return NATIVE_IMPORTER(name, globals, locals, fromlist, level)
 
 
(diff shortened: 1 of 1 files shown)


mdipierro:
This cannot be right: if('gluon' not in name and not name.startswith('.')
what if one does import agluon or import gluonic? Also it does not seem to do what advertised. Let's discuss it on the dev group.

mdipierro:
Also... this:
   from dal import DAL
should not work because gluon is not in path. this:
   from gluon.dal import DAL
should work and (possibly) import the same class as from gluon import DAL.

ilvalle:
Hi
2015-01-16 15:51 GMT+01:00 mdipierro <notifications@github.com>:

> This cannot be right: if('gluon' not in name and not name.startswith('.')
> what if one does import agluon or import gluonic?
>
Good catch, I should write a less permissive test

> Also it does not seem to do what advertised. Let's discuss it on the dev
> group.
>
That sounds strange to me, have you got an example?
An open discussion is here
https://groups.google.com/d/topic/web2py-developers/PGcEFWMqnRc/discussion

gluon/custom_import.py
```diff
@@ -67,6 +67,13 @@ def custom_importer(name, globals=None, locals=None, fromlist=None, level=-1):
         import_tb = None
         try:
             try:
+                if ('gluon' not in name and not name.startswith('.') and
+                   'direct_import' not in locals):
+                    try:  # Try to load a gluon class
+                        oname = 'gluon.%s' % name
+                        return NATIVE_IMPORTER(oname, globals, locals, fromlist, level)
+                    except:
+                        pass
```
dmvieira: What kind of exception?

gluon/custom_import.py
```diff
@@ -67,6 +67,13 @@ def custom_importer(name, globals=None, locals=None, fromlist=None, level=-1):
         import_tb = None
         try:
             try:
+                if ('gluon' not in name and not name.startswith('.') and
```
mdipierro: I still do not think this is a good idea.

dmvieira: +1

ilvalle: What do you propose to fix the mentioned issues?

 Paolo

2015-01-23 21:25 GMT+01:00 Diogo <notifications@github.com>:

dmvieira: Perhaps removing 'gluon.' from name and trying to import. If you just compare if exists gluon in name we will get more bugs... :(

mdipierro: I this if name does not start with 'gluon.' and does not start with '.' should try name = 'gluon.'+name but we should not check for "gluon" in name.

dmvieira: You can save one condition: if name.split('.')[0] == 'gluon' ... Something
like this...
Em 23/01/2015 18:59, "mdipierro" <notifications@github.com> escreveu:

mdipierro:
Do we still need something like this?

ilvalle:
It basically allows what's mentioned in the first post.

mdipierro:
No longer necessary:
>>> from gluon import DAL
>>> from gluon.dal import DAL as d
>>> d is DAL
True

PR #593, opened 2015-01-20
Issue/2035: cache to support filenames greater than 256 chars on Windows

kirsn:


leonelcamara:
Can you change this to a simple + instead of using a join and a list?  
    
    PS C:\Users\Leonel> python -m timeit "dst=''.join(['foo', 'bar'])"  
    1000000 loops, best of 3: 0.231 usec per loop  
  
Replacing list with tuple yields a 50% speedup:
  
    PS C:\Users\Leonel> python -m timeit "dst=''.join(('foo', 'bar'))"
    10000000 loops, best of 3: 0.111 usec per loop  
  
Simply using the + operator thus avoiding the join function call:  
  
    PS C:\Users\Leonel> python -m timeit "dst='foo' + 'bar'"
    100000000 loops, best of 3: 0.0202 usec per loop

Conclusion using + is more than 11x faster in this scenario.

kirsn:
Leonel, thanks for letting me know about this. Will remove this PR, and issue another with the changes.
I settled for a "join", thanks to this link: http://waymoot.org/home/python_string/ . Was lazy and didn't verify the results, or notice that it is from 2004 :-/

PR #596, opened 2015-01-21
Issue 2035: Fix the caching on Windows

kirsn:
Issue 2035: Fix the caching on Windows - issue with filenames greater than 256 chars"


leonelcamara:
Looks good!

PR #598, opened 2015-01-23
Script for centos7 now also installs 'unzip' package

dmatic:
Some virtual server providers (like DigitalOcean) do not have unzip
installed in their centos7 image so we need to install it in our script.


mdipierro:
Thanks dmatic

PR #601, opened 2015-01-25
Bug fix SQLFORM.factory

nursix:
...was crashing with a NameError. Import at the module level doesn't seem to work.

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -1733,6 +1733,7 @@ class SQLFORM(FORM):
         if 'table_name' in attributes:
             del attributes['table_name']
 
+        from gluon import DAL
         return SQLFORM(DAL(None).define_table(table_name, *fields),
                        **attributes)
 


niphlod:
uhm. strange. DAL should be imported from gluon.dal and work without issues. Let me try that....

niphlod:
thanks @nursix , but is seems a small part of a larger issue. 
fixing the larger should fix also this. 
@mdipierro : don't merge this for now, @gi0baro is looking into it.

gi0baro:
@nursix do you know if it was working on R-2.9.12 and it's a master related problem only?

nursix:
Uhm...you can import from either gluon or gluon.dal - both works as long as you import it inside factory(). Importing at the module level causes an ImportError due to a circular dependency.

Seems that Massimo has already committed this fix to trunk, so my PR may be obsolete now.

nursix:
https://github.com/web2py/web2py/commit/554722318a8e56241064a2c4248c6eeeced72400

...is basically the same fix. So, closing this one.

PR #758, opened 2015-01-26
we should always avoid importing "optionals" without a try:except.

niphlod:
one clear case of "should never happen"... that happened.
in this case docs were failing builds on readthedocs because of 
pypyodbc not being able to load any odbc library.


niphlod:
BTW: the current implementation fails even on latest *buntu

PR #762, opened 2015-01-27
updated Pg8000

ilvalle:


niphlod:
sys.path is getting weird.....

ilvalle:
I see but it's the only way I found so far

niphlod:
let's rethink a bit in web2py-developers.

gi0baro:
@ilvalle @niphlod why do we need sys.path hack? from .contrib import should work..

ilvalle:
I tried a bit more by removing all references of pg8000 in all imports and eventually the import worked.
The bad part is that, the driver doesn't work, namely it raises exceptions when make references to pg8000.*
for example here: https://github.com/ilvalle/web2py/blob/pg8000/gluon/contrib/pg8000/__init__.py#L327
I don't think we should dive into all this details.
I hope @mfenniak, the pg8000 maintainer, may have a better solution for that.

gi0baro:
Ok.. so maybe we can accept the sys.path as a temporary workaround waiting for better solutions.
@niphlod what do you think?

niphlod:
the point now is : why web2py should work around something that is so buggy ? let's wait for a proper fix and live without tampering web2py internals.

mfenniak:
Hey all.  If I'm understanding the discussion here, it seems like you're having troubles incorporating pg8000 into this codebase because of the use of absolute imports in the code?  It would definitely make sense to change pg8000 to use relative imports for it's cross-module references, and I'd happily accept a PR to fix that issue.  If a PR isn't forthcoming, I'd fix it myself, but I can't do that immediately.  The only concern I have is that pg8000 supports older versions of Python back to 2.5, but I believe that's when relative imports were first added to Python so it shouldn't be a problem.

ilvalle:
@niphlod, @gi0baro I understand your points. I see two options:
1. keep the pg8000 (3years old) we have now in contrib and wait for a newer pg8000
2. integrate the current pg8000 (the one which requires sys.path to be updated) and update it as soon as  a fix, for relative imports, is provided (by @mfenniak or anyone else).
My proposal is go with 2, not only for the updated driver but also because I don't see so critical to live *for a while* with 'contrib' in sys.path.

gi0baro:
@ilvalle I agree for no. 2
But I think we should update this to remove sys.path hack before the next release

gi0baro:
@ilvalle since https://github.com/mfenniak/pg8000/issues/71 is merged, you can update pg8000 to latest code revision and remove the `sys.path` hack ;)

ilvalle:
Ok, I've updated pg8000 to the last version and rebase the last commit of this PR. 
Now it should be fine to be merged.

mdipierro:
Thank you all for the hard work!

PR #771, opened 2015-02-03
Fix for map_hyphen breaking codemirror editor

### LINKED ISSUE #769: Setting map_hyphen=True in routes.py breaks codemirror editor
The codemirror breaks once I set the `map_hypen` to `True`.

In my `routes.py` I have:
```python
routers = dict(
    BASE=dict(
        default_application='mycoolapp',
        map_hyphen=True,
    )
)
```

When I try to use the web-based web2py editor it fails to load. 

Looking at the console I see the following 404's:

screen shot 2015-02-02 at 5 35 16 pm

The path goes from being an _2.9.5 to -2.9.5 which is expected behavior for the URLS, however because the directory starts with a _ and gets changed to - it causes problem. For example http://localhost:8000/admin/static/**-2.9.5**/js/typeahead.min.js

erikmontes:
Fixes #769

### DIFF
--- a/gluon/rewrite.py
+++ b/gluon/rewrite.py
@@ -1257,9 +1257,9 @@ class MapUrlOut(object):
         "Builds a/c/f from components"
         acf = ''
         if self.map_hyphen:
-            self.controller = self.controller.replace('_', '-')
             if self.controller != 'static' and not self.controller.startswith('static/'):
                 self.application = self.application.replace('_', '-')
+                self.controller = self.controller.replace('_', '-')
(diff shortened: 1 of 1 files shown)


niphlod:
uhm. I'm not convinced this is the right fix.... and in fact I can't reproduce the original issue.......
my "_2.9.5 to -2.9.5" doesn't take place in 2.9.12. The problem is already fixed by https://github.com/web2py/web2py/commit/c0536d3b7487e73599b1f2ae1fcb963b91d41e8f . Can you please doublecheck ?

PR #775, opened 2015-02-05
Fix enable element

leonelcamara:
enableElement was doing this:

    var method = el.is('button') ? 'val' : 'html';

instead of:

    var method = el.is('button') ? 'html' : 'val';

### DIFF
--- a/applications/welcome/static/js/web2py.js
+++ b/applications/welcome/static/js/web2py.js
@@ -515,7 +515,7 @@
 
       /* restore element to its original state which was disabled by 'disableElement' above*/
       enableElement: function(el) {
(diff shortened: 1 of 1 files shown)


niphlod:
uhm, I recall fixing the same line for a different reason.... sure it's working all fine ? what was not working before ?

leonelcamara:
I was using disableElement/enableElement on a button I have, and it lost the text it had there and it had a new value attribute. The reason was enableElement was using jquery val for the button instead of html. Basically if the button is an input you want to use val and you want to use html if it's a button. disableElement was already doing the right thing whereas enableElement wasn't.

PR #776, opened 2015-02-05
Added CAPTION to gluon/html.py as it is often used with tables. 

gorosz:
Also included test for new tag in gluon/tests/test_html.py

### DIFF
--- a/gluon/html.py
+++ b/gluon/html.py
@@ -96,6 +96,7 @@ __all__ = [
     'SPAN',
     'STRONG',
     'STYLE',
+    'CAPTION',
     'TABLE',
     'TAG',
     'TD',
@@ -1766,6 +1767,11 @@ class COLGROUP(DIV):
     tag = 'colgroup'
 
 
+class CAPTION(DIV):
+
+    tag = 'caption'
+
+
 class TABLE(DIV):
     """
     TABLE Component.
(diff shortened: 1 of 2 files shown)


niphlod:
we have a strict policy to avoid cluttering the global namespace.......... use TAG.caption() instead...

gorosz:
The problem whit that is it wrap the caption into a new tbody tag which can be problematic with some javascript/jquery applications for example dataTables

niphlod:
what do you solve with CAPTION that you can't with TAG.caption()?

gorosz:
If I define the CAPTION class I can add it to the TABLE's _fixup method so it is not wrapped in a TR and the html rendering engine does not create a new tbody for that TR element.
If it is possible with TAG.caption() it is absolutely fine.

leonelcamara:
Gorosz has a point. Maybe add a test to make sure a caption in a table doesn't get wrapped.
  
We should probably just remove automatic wrapping (backwards incompatible sadly) otherwise this is the only way I see to solve it.

mdipierro:
The new helpers in Form.py do not do automatic wrapping. So web3py will not do it.

mdipierro:
One solution here is to change the behavor of TABLE fixup. I will do so.

mdipierro:
I do not want to bloat this. So no for now.

PR #781, opened 2015-02-07
Update/codemirror 4.12

Pivert:
Hi, 

This is just a codemirror udpate to 4.12. It includes the Fix for the issue 2046 as well.
I removed most of the unused languages to keep it as small as possible.

François Delpierre


Pivert:
Some problems with ctrl+s and the :w that are not properly taken. But it was already the case with the previous version. It's really missing not not bind at least the :w, wq, bn, bp, bf, fl, e! to better use the tabs and save in Vim mode.

ilvalle:
You've included a lot of codemirror files unused in web2py (Addon, html example ...).

Pivert:
I removed the most space consuming part that was taken by all the supported languages.
I'll check if I can remove some more stuff but it might take up to the next w-e.

ilvalle:
Start by removing all addons unused by the online editor, then all *.html and plain files (i.e. readme, contributing and so on)

PR #795, opened 2015-02-14
removed unsed codemirror files

ilvalle:
It should be fine to remove those files


ilvalle:
Double check that everything is fine

PR #800, opened 2015-02-18
Added the 2 missing files. This is probably consequences of codemirror

Pivert:
Bug fix.
The first file was requested when Creating a new file.
The second is a css used by codemirror in non vim mode.

Regards,

PR #802, opened 2015-02-18
Issue/801 - Fix the :write not saving problem in codemirror vim mode

Pivert:


ilvalle:
@Pivert one of your commit is in both PRs. 
Mind that usually you should create a new branch from the master one, not from an other branch, otherwise the new branch will inherit the commit of the other one.
Anyway, closing https://github.com/web2py/web2py/pull/800 is more than fine.

Pivert:
@ilvalle Thanks Paolo, just realized it yesterday when making the second pull request. Hopefully it was just 2 added files. I need to be a bit more comfortable with git in shared projects. This should not happen anymore.

PR #821, opened 2015-02-25
Fixed issue 819

leonelcamara:
The problem was that it was only checking for button to use html instead of val, what you need is to only use val if it it's an input and use html for everything else.

### DIFF
--- a/applications/welcome/static/js/web2py.js
+++ b/applications/welcome/static/js/web2py.js
@@ -490,7 +490,7 @@
        * and prevent clicking on it */
       disableElement: function(el) {
         el.addClass('disabled');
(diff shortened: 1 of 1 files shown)


niphlod:
is this reverting the EXACT same lines you modified already in https://github.com/web2py/web2py/commit/53a774827e38c6e5bb3bc32466625fe663d7b537 ?

leonelcamara:
Not exactly. That fix was correct, considering the previous logic was checking if it was a button. The problem here was that it really shouldn't test for a button, it should test for input and user jquery.html for everything else.  This way it will work for any html element.

 I think this testing for button is an historic thing coming from a time where this was only used to disable form buttons which were either an input or a button.

PR #827, opened 2015-03-02
fix sessions2trash bug, auth not exist in single_loop

KabluBR:


niphlod:
how were you using single_loop()? it's supposed to be run in the app environment where auth is there....

niphlod:
moreover, this doesn't seem to honor the -x argument......... please tell us what is the original bug.

KabluBR:
As you can see in the description at the top of the file:
 # Delete session in a module (move to the modules folder)

In modules there is no auth set,  you need to set auth using current.

Anyway if you use current will work both ways... in modules or not...

niphlod:
nope, it won't. your commits aren't taking into consideration specifying the -x option

niphlod:
nope ^ 2, I stand corrected (fetched the wrong patch). This works as advertised.

KabluBR:
per my understanding you are talking about calling by cron or manually by command line...
see that I am talking about calling from inside app using modules...

and I cannot see where the commit is not taking -x option into consideration, if it is looking for expiration parameter.

def single_loop(expiration=None, force=False, verbose=False):
if expiration is None:  **************************** <<--------------------------------
  if current.session.auth:
    expiration = current.session.auth.settings.expiration

niphlod:
I'm spinning my head around because I verified at first that it was not working........what I did was calling the method inside the app, but I had no session ! This fails badly (though I see the original idea) because there could be places in your app where session.auth is not there at all.

The best option is to just rearrange the docs at the top and propose

    def delete_sessions():
        single_loop(auth.settings.expiration)

KabluBR:
yes it works

or you add in the if:  'session' in current

;)

niphlod:
what works ?

KabluBR:
your idea... ;)

niphlod:
ok. so can I close this PR and you'll make another one just fixing the docs at the top ?

niphlod:
ok,  I'll do it.

PR #846, opened 2015-03-10
caches recursive selects for references

niphlod:
Why didn't we think before ? References are lazy, but when asked 
for representation why should we fetch the same referenced record
over and over ?

NB: the code caches only represent() fetching records from tables, that
are the only idempotent represent() calls out there. We could cache
every representation only asking for every represent() call to be 
idempotent. As I feel there will be someone returning random values,
I left it out.

I'll make a PR soon for pyDAL to fix the same behaviour.


mdipierro:
Nice!

PR #853, opened 2015-03-16
Improve PEP8 gluon/languages.py

BuhtigithuB:


BuhtigithuB:
Travis Fails don't seem to coming from the patch...

PR #869, opened 2015-03-23
Improve PEP8 gluon/settings.py

BuhtigithuB:


BuhtigithuB:
After this one there is only 5 remaing!!

:)

PR #873, opened 2015-03-24
Update ajax_examples.py so it does what it says.

sp1d3rx:
Fixed the example so it actually does what it says. It said "The last 10 entries will appear sorted in a table below.". What it did was if the number of entries was 10, it would truncate the entries and start over again. So, in order for this to show the last 10 entries, it has to delete the oldest entry after 10. The best way to do that is to not sort the list, except when displaying and then delete the first entry in the list before inserting the 11th entry.

### DIFF
--- a/applications/examples/controllers/ajax_examples.py
+++ b/applications/examples/controllers/ajax_examples.py
@@ -3,12 +3,13 @@ def index():
 
 
 def data():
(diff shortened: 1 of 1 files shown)


sp1d3rx:
I just hate when examples and tutorials don't do what they purport to do.

PR #877, opened 2015-03-26
make unique_key unique also for the database table definition

michele-comitini:
This should be a big speed improvement.  I wonder why it was not unique before.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -890,7 +890,7 @@ class Session(Storage):
                     Field('created_datetime', 'datetime',
                           default=request.now),
                     Field('modified_datetime', 'datetime'),
-                    Field('unique_key', length=64),
+                    Field('unique_key', length=64, unique=True),
                     Field('session_data', 'blob'),
                     migrate=table_migrate,
                 )


niphlod:
sessions are looked up by the id, that's already indexed by default.

michele-comitini:
Then is this wrong?
https://github.com/michele-comitini/web2py/blob/session_unique_key_must_be_UNIQUE/gluon/globals.py#L909

2015-03-26 15:24 GMT+01:00 niphlod <notifications@github.com>:

niphlod:
no, it's not. That is translated to a 

SELECT blablabla from web2py_session_table where id = something AND key = something

any query optimizer will choose the id pkey (that has the added benefit of being an integer, and always ordered as the records are on the disk) as the one to filter with.
enforcing a unique constraint is not wrong-per-se (model-wise), but I'm not sure indexing an uuid will bring more speed to the table. unique indexes are not clustered and stored "on another section of the disk", and are usually translated to an ordered set of "something" that have "pointers" to the pkeys (or OIDs, if you're more "postgresql-minded"). 
Maintaining a unique index on a uuid is probably one of the things that is harder to do for a backend.

PR #888, opened 2015-03-31
Improve PEP8 gluon/utils.py

BuhtigithuB:


BuhtigithuB:
You have them all finally!!

I realise what was the problem... When I proceed to the PEP8 corrections all my branches were deriving from one another... There were only one leaf of master, so all my fixes was including the precedents fix and I can't make standalone PR for each of them... Simone help me understand the leaf concept of git and how to make sure my branch and fix has the last master HEAD as base...

:)

Happy merging...

PR #900, opened 2015-04-04
added basic .gitignore for web2py applications

ilvalle:


mdipierro:
So what do we do with private? We suggest people git commit it? There is a reason it is called private. But now we have config.ini in there and the app does not work without. Ouch!

KabluBR:
Maybe we can use private/appconfig.ini to store all config but secrets...
And create private/secrets.ini to store only passwords, and this one will be ignored by .gigignore...
Also we can keep some secrets_sample.ini...

niphlod:
Imho this is wrong. The databases folder should not be excluded in all cases. For the "private" part of the conversation, we had credentials stored in models/Db.py: let the people choose their path, don't enforce anything of this

ilvalle:
I've added private folder to gitignore.
For what concern config.ini I think it should be placed somewhere else.

mdipierro:
Moving it now will cause upgrade problems.

KabluBR:
I think there is no advantage using external ini file to configure app if not used to keep secrets / passwords in a safe place "private folder".

The big advantage of using private is to keep the secrets out of versioning systems or backups.

Or it will just be more script/files to load, more coding, more complexity, more iops on disk...

mdipierro:
I am not convinced this belongs in web2py. If people want to use git with apps, they can write their own .gitignore. We should not put the cart in front of the horses. If anything we should allow the web2py admin interface to support git and then we should add .gitignore.

ilvalle:
New users don't know the structure of a web2py application. Given that, if
they start using git, they will fall in the problem of writing a
.gitignore, which is not so obvious in the beginning of the development.
We can provide it either directly in the app or through other indirect
means (admin interface, recipe on web2pyslices and so on...).

 Paolo

2015-04-05 23:06 GMT+02:00 mdipierro <notifications@github.com>:

niphlod:
1) not everyone wants to use git
2) there are some patterns excluded that works probably just for your case but not everybody's
3) let the users deal with .gitignore in their own app

ilvalle:
I think this is a missing feature for new web2py users.

PR #913, opened 2015-04-13
Update/gluon contrib pypyodbc

BuhtigithuB:


BuhtigithuB:
I think there is something wrong :(

PR #915, opened 2015-04-13
Update gluon/contrib/pypyodbc.py 1.3.0 -> 1.3.3

BuhtigithuB:


BuhtigithuB:
Here we go!

On my first try my fork repo were not even with web2py official repo... fetch and merge upstream/master when wrong, I guess that was the reason... I reset --hard and fetch merge and reapply the update and now it works.

PR #929, opened 2015-04-19
Locale files for calendar.js

leonardopf:
I think that this is the best aproach. As using T would not be able to change the weekends.

The problem that might be here is this:
response.files.insert(3,URL('static','js/calendar/%s.js' % T.accepted_language or 'en'))
response.files.insert(4,URL('static','js/web2py.js'))

if a new language be created, it will not be replicated to /js/calendar/lang.js.


niphlod:
this is a gigantic mess AND it doesn't solve the problem: all files you added are still in english.

mdipierro:
I think this is beyond the scope of web2py. We cannot include all these files in scaffolding. Sorry.

PR #936, opened 2015-04-21
added newer Recaptcha2 class to deal with v2.0. Fixes #919

### LINKED ISSUE #919: Support the new reCaptchas
Are there any plans to support the new "I'm not a robot" reCaptchas out of the box? I know the old styled recaptchas are currently supported, but it would be better to use the newest.

https://developers.google.com/recaptcha/

niphlod:
Improvements over the "old" v1.0
- behaves well also without javascript
- use_ssl is redundant, v2.0 works only in https mode
- ajax is not useful anymore as the newer API is a lot easier

Adjusted also the addrow() method that was missing newer formstyles.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -60,7 +60,7 @@ except ImportError:
         # fallback to pure-Python module
         import gluon.contrib.simplejson as json_parser
 
-__all__ = ['Mail', 'Auth', 'Recaptcha', 'Crud', 'Service', 'Wiki',
+__all__ = ['Mail', 'Auth', 'Recaptcha', 'Recaptcha2', 'Crud', 'Service', 'Wiki',
            'PluginManager', 'fetch', 'geocode', 'reverse_geocode', 'prettydate']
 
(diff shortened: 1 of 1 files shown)


mdipierro:
Do we have to call this Recaptcha2? can we replace the old one?

niphlod:
widgets dimensions aren't the same: it'll break existing designs, although it won't break functionality. Moreover, arguments for the old one don't match the newer. We can recommend this over the old, but not replace it.

mdipierro:
Can we expose a single interface that renders as one or the other depending on the arguments?

niphlod:
Don't think so, sorry but they're too different.

mdipierro:
OK. But eventually we should break tools into sub modules.

niphlod:
you're preaching to a quire. Extrapolating recaptcha and recaptcha2 is by far the easiest sub-task in the great project of breaking tools.py.

mdipierro:
I think each class in tools is its own universe. They do not depend on each other.

PR #950, opened 2015-05-02
Fix List behaviour and added new feature

cassiobotaro:
Now List has the behavior equal your comments(look at __getitem__) and we can do:
test = request.args(0 , otherwise=URL('default', 'index')) and i will be redirected if args is not found.
Otherwise parameter is not binded with cast.

### DIFF
--- a/gluon/storage.py
+++ b/gluon/storage.py
@@ -298,6 +298,7 @@ class List(list):
             if not value and otherwise:
                 raise ValueError('Otherwise will raised.')
         except (ValueError, TypeError):
+            from http import HTTP, redirect
             if otherwise is None:
                 raise HTTP(404)
             elif isinstance(otherwise, str):
--- a/gluon/storage.py
+++ b/gluon/storage.py
@@ -269,53 +269,50 @@ class FastStorage(dict):
 
 
 class List(list):
+
     """
     Like a regular python list but a[i] if i is out of bounds returns None
-    instead of `IndexOutOfBounds`
+    instead of `IndexOutOfBounds`.
     """
 
-    def __call__(self, i, default=DEFAULT, cast=None, otherwise=None):
+    def __call__(self, i, default=None, cast=None, otherwise=None):
         """Allows to use a special syntax for fast-check of `request.args()`
         validity
-
         Args:
             i: index
             default: use this value if arg not found
             cast: type cast
             otherwise: can be:
-
              - None: results in a 404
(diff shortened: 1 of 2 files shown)


gluon/storage.py
```diff
@@ -269,54 +269,52 @@ def update(self, *args, **kwargs):
...
-        n = len(self)
-        if 0 <= i < n or -n <= i < 0:
-            value = self[i]
-        elif default is DEFAULT:
-            value = None
-        else:
-            value, cast = default, False
-        if cast:
-            try:
+        value = self[i] or default
```
mdipierro: This breaks if self[i] is out of bounds.

gluon/storage.py
```diff
@@ -269,54 +269,52 @@ def update(self, *args, **kwargs):
...
-                from http import HTTP, redirect
-                if otherwise is None:
-                    raise HTTP(404)
-                elif isinstance(otherwise, str):
-                    redirect(otherwise)
-                elif callable(otherwise):
-                    return otherwise()
-                else:
-                    raise RuntimeError("invalid otherwise")
+            if not value and otherwise:
```
mdipierro: The ability to specify cast=int and default=None was a feature, not a bug.

mdipierro:
It is not clear to me why the need to the change. It seems to break a lot of use cases.

gluon/storage.py
cassiobotaro: Don't break because of modification of __getitem__ function , see line 312

cassiobotaro:
Don't break and to prove that as you can see, i added more tests. I preserve the old behaviour, but now if you can do somethink like request.args(0, otherwise=redirect(URL('default', 'index'))).
Previously the otherwise only will be considered if you have cast, now both works.

cassiobotaro:
Another point is, the doc says "Like a regular python list but a[i] if i is out of bounds returns None
-    instead of `IndexOutOfBounds`	" but this is not True. a(i) returns None, but a[i] will raise an exception.
getitem method was rewrite to to do this now.

gluon/storage.py
cassiobotaro: I think that i don't modify it.

mdipierro:
Before:
In [1]: print request.args(0,cast=int,default=None)
None

After:
In [1]: print request.args(0,cast=int,default=None)
HTTP: 404 NOT FOUND

This is incorrect. Will revert.

Issue #955, opened 2015-05-04
gluon.storage.List does not support reverse slicing (i.e. with negative step)


nursix:
Unlike normal lists, gluon.storage.List instances do not support slicing with negative step like [::-1]:

Normal list:

    >>> x = [1,2,3]
    >>> print x[::-1]
    [3, 2, 1]

gluon.storage.List:

    >>> x = List([1,2,3])
    >>> print x[::-1]
    None

This breaks our code, unfortunately :/

nursix:
In fact, it doesn't support step at all:

    >>> x = List([1,2,3,4])
    >>> print x[1:4]
    [2, 3, 4]
    >>> print x[1:4:2]
    None

The problem is probably that __getitem__ doesn't check whether i is a Slice - but simply assumes it's a simple index.

nursix:
Our problem is btw with request.args - we're not using gluon.storage.List otherwise. But we would like to step (backwards) through request.args, and in earlier versions it did behave like a normal list - now that has changed.

nursix:
Please allow me also to say that for request.args the gluon.storage.List behavior is rather confusing and obscuring defects.

Most of our code expects request.args to indeed behave like a list and raise an IndexError when the index is out of bounds - but where it returns None, this would lead to TypeErrors in later passages instead, which are hard to trace back to the actual error.

We have already banned Storage for this exotic behavor - it's much harder to debug these "convenience" objects because they do not raise exceptions like the objects they replace - and for most developers it is totally unexpected that request.args[x] would give None under any circumstances rather than raising an IndexError.

I'm not convinced that this is actually an improvement - I think it makes it less safe and harder to debug.

cassiobotaro:
About slice, you are right, i added these tests  and i fix early as possible. But about args[x], docs say: " Like a regular python list but a[i] if i is out of bounds returns None instead of `IndexOutOfBounds`."   Then i don't think it's unexpected.
But you have a interesting point:

"
We have already banned Storage for this exotic behavor - it's much harder to debug these "convenience" objects because they do not raise exceptions like the objects they replace - and for most developers it is totally unexpected that request.args[x] would give None under any circumstances rather than raising an IndexError.

I'm not convinced that this is actually an improvement - I think it makes it less safe and harder to debug.
"

Let's discuss that.

nursix:
Ok, maybe I missed the updated docs about request.args. We've been using web2py for many years, and at some point I've just stopped reading it.

[...]

PR #956, opened 2015-05-04
More changes in List (request.args)

cassiobotaro:
- Better documented List
- Otherwise not binded with cast
- new tests to verify old behaviour 
- test otherwise without cast and defaut
- verify if behave like a list 
- more test with call function
Related to #955

### DIFF
--- a/gluon/storage.py
+++ b/gluon/storage.py
@@ -22,8 +22,7 @@ import gluon.portalocker as portalocker
 __all__ = ['List', 'Storage', 'Settings', 'Messages',
            'StorageList', 'load_storage', 'save_storage']
 
-def DEFAULT():
-    return 0
+DEFAULT = lambda: 0
 
 
 class Storage(dict):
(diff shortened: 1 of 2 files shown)


cassiobotaro:
@nursix PR done, if you have another error, please report. Many thanksany thanks for all.

cassiobotaro:
@mdipierro don't forget this commit before 2.10.5 to avoid problems.

nursix:
Ah - ok, I see that this hasn't yet been committed to trunk (see my comment https://github.com/web2py/web2py/commit/c36c39178610867330ebf1927a2a3edeb7dcb35e#commitcomment-11104745).

Sorry for bothering you again, thanks for fixing it. Waiting for this to get merged then...

mdipierro:
I will take this pull request but please reinstate DEFAULT = lambda: 0. it is kind of a convention everywhere in web2py.

PR #985, opened 2015-05-29
Invite user

peregrinius:
Invite by email another user to access your application. Note, my
initial version was built on Auth.register_bare which doesn't seem to be
in this repository???


mdipierro:
This is a great idea. I will merge then do some minor re-factoring.

PR #1046, opened 2015-08-18
added default HttpOnly cookies

ShySec:
https://www.owasp.org/index.php/HTTPOnly#What_is_HttpOnly.3F

supports "session._js_cookies = True" to restore old behavior

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -1029,7 +1029,7 @@ class Session(Storage):
         if self._forget:
             del rcookies[response.session_id_name]
             return
-        if not self._js_cookies:
+        if self.get('httponly_cookies',True):
             scookies['HttpOnly'] = True
         if self._secure:
             scookies['secure'] = True
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -1023,10 +1023,16 @@ class Session(Storage):
     def _fixup_before_save(self):
         response = current.response
         rcookies = response.cookies
-        if self._forget and response.session_id_name in rcookies:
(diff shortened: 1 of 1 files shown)


niphlod:
uhm. "_js_cookies" doesn't really sound as what it provides. Maybe httponly_cookies is best. And I'd go for issuing httponly cookies by default without the option to remove the httponly part

ShySec:
I agree with default-secure options that protect users. At the same time, I agree with supporting non-standard configurations without modifying web2py source code. The new configuration variable (session.httponly_cookies) would need to be deliberately set in all sessions, which should make the setting better scoped and more obvious than web2py source code modification (and only slightly harder).

niphlod:
this wasnt't ready to merge!

mdipierro:
I thought he made the changes you requested.

niphlod:
uhm. a part from the fact that we rarely touch globals without 31290 cross-checks, have someone tested it ?

mdipierro:
reverted

ShySec:
I don't know what 31290 references. How do we move this forward?

niphlod:
adding tests, mainly. 
The "anti-rush" was because a new version was going to be released (2.12.3) which was supposed to be just a fix for an incorrect pydal inclusion. 
We can move with the usual schedule here, but regression tests IMHO should be added to avoid regression(s) in a piece of code that gets executed at EVERY request.

ShySec:
It doesn't appear I can advance this PR directly; submitting a new one with all updates + tests.

Issue #1048, opened 2015-08-19
SQLFORM formstyle overridden when validation fails


timnyborg:
A portion of SQLFORM.accepts() overrides widgets that have been explicitly set in the DAL and styled with a formstyle, when the form has failed validation.  I do not understand what the point of the for-loop that does this is, and I've commented it out in my source with no problems (so far), which fixes the styling issue.

```
    elif (not ret) and (not auch):
        # auch is true when user tries to delete a record
        # that does not pass validation, yet it should be deleted
        for fieldname in self.fields:

            field = (self.table[fieldname]
                     if fieldname in self.table.fields
                     else self.extra_fields[fieldname])
            # this is a workaround! widgets should always have default not None!
            if not field.widget and field.type.startswith('list:') and \
                    not OptionsWidget.has_options(field):
                field.widget = self.widgets.list.widget
            if field.widget and fieldname in request_vars:
                if fieldname in self.request_vars:
                    value = self.request_vars[fieldname]
                elif self.record:
                    value = self.record[fieldname]
                else:
                    value = field.default
                row_id = '%s_%s%s' % (
                    self.table, fieldname, SQLFORM.ID_ROW_SUFFIX)
                widget = field.widget(field, value)
                parent = self.field_parent[row_id]
                if parent:
                    parent.components = [widget]
                    if self.errors.get(fieldname):
                        parent._traverse(False, hideerror)
                self.custom.widget[fieldname] = widget
        self.accepted = ret
        return ret
```

In my case:
Field('string1') keeps its bootstrap styling
Field('string2', widget=SQLFORM.widgets.string.widget) loses its bootstrap styling (as does any field with widget=...)

It breaks the styling because formstyle is run during FORM.**init**(), which always occurs before FORM.accepts()

niphlod:
can we close this @timnyborg given that #1049 has been merged ?

timnyborg:
#1049 wasn't merged, so this is still awaiting a solution from @mdipierro, unless I can figure something out

mdipierro:
This is hard to fix in current form logic. Will be addressed in new forms.

timnyborg:
This continues to a be a problem, which makes it hard to customize widgets in forms.  Any possibility of getting it fixed?

timnyborg:
If the point of this (see Massimo's comment on https://github.com/web2py/web2py/pull/1049) is to prevent list:string from getting wiped out, can we not change the following?

`if field.widget and fieldname in request_vars:`
to
[...]

PR #1049, opened 2015-08-19
Update sqlhtml.py - fix failed validation removing styling

timnyborg:
Addresses https://github.com/web2py/web2py/issues/1048, and works in my application.

However, I don't know what the purpose of the existing code is: why does it override widgets, and why only those which are explicitly set?  Isn't this just replicating work done in __init__()?  What could this break?

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -1515,31 +1515,6 @@ class SQLFORM(FORM):
         elif (not ret) and (not auch):
             # auch is true when user tries to delete a record
             # that does not pass validation, yet it should be deleted
-            for fieldname in self.fields:
-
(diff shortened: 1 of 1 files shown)


mdipierro:
I agree this is broken but that code serves a purpose and can be improved but cannot be removed. The purpose is the following. If you have a table with a list:string field and the SQLFROM does not pass validation (but not because of the list:string), without that code the list:string is wiped out upon submission. That makes sure that all fields (except upload fields) remain populated when the form is re-rendered after a failed validation. The logic does not take into account customizations made to the widgets, for example bs3 classes and indeed needs to be fixed. I may be able to work on it in the next few days.

timnyborg:
Thanks for the explanation

PR #1051, opened 2015-08-19
added HttpOnly cookies (default)

ShySec:
added unit tests for cookie layout, secure cookies, and HttpOnly cookies
Session.httponly_cookies=False to revert HttpOnly cookies

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -1023,10 +1023,16 @@ class Session(Storage):
     def _fixup_before_save(self):
         response = current.response
         rcookies = response.cookies
-        if self._forget and response.session_id_name in rcookies:
+        scookies = rcookies.get(response.session_id_name)
+        if not scookies:
+            return
+        if self._forget:
             del rcookies[response.session_id_name]
-        elif self._secure and response.session_id_name in rcookies:
-            rcookies[response.session_id_name]['secure'] = True
+            return
+        if self.get('httponly_cookies',True):
+            scookies['HttpOnly'] = True
+        if self._secure:
+            scookies['secure'] = True
 
     def clear_session_cookies(self):
         request = current.request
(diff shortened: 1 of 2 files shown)


ShySec:
https://www.owasp.org/index.php/HTTPOnly#What_is_HttpOnly.3F

ShySec:
previous PR https://github.com/web2py/web2py/pull/1046

mdipierro:
Let's wait for niphlod's comment than I will review it. ;-)

ShySec:
Ping. I believe this update meets all niphlod's previous concerns (namely applying post-2.12.3-release and the addition of regression tests).

niphlod:
pong, @niphlod was on holidays, but planning for a big return on the scenes. 
Joking (on the *big* return), but hey, it's been 5 days, not a month...

BTW: tests for the new feature are there, but that piece of code gets called for every kind of session, file, db or cookie based. Frankly, on 4 or 5 times I got my hands dirty on globals.py it resulted in a mess. 
It was since rewritten basically from the ground up to eliminate the complications, but the result was not really __that__ linear. That's why I tread lightly  in regards of this module.

BTW2: @mdipierro with all our "OWASP-fixer-friends", why didn't this come up earlier ? :D

BTW3: I assume everyone tested components and ajax with this already. From the specs, adding httponly in a modern browser affects only "reading document.cookie", that returns an empty string...but $.ajax() should be allowed to copy the cookie and send it back to the server ... but one can't ever **be sure** of "specs" without testing.

ShySec:
I did test components and AJAX. For clarity sake, jQuery ($) cannot access or copy the cookie, but $.ajax will include the HttpOnly cookie (provided by the browser).

mdipierro:
I reviewed this code and it looks good to me. The rewrite does not change behavior. The change in functionality is optional and comes with tests. It addresses a potential vulnerability. Good job.

PR #1060, opened 2015-09-03
Closes #1059: get `response.view` from the environment

### LINKED ISSUE #1059: `response.render()` uses the original `response.view`
The problem
-----

When using `response.render()`, Web2py works fine for the first request. But it uses wrong view name in subsequent requests.

My same code worked in v2.9.11.

The scenario
-----

I use views to render parts of my page in this way:

```
def my_action():
    my_div = response.render('div_layout.html')
    # more code...
    return dict(my_div=my_div, other_vars=...)
```

Web2py must use these views:
  1. `div_layout.html` in the explicit `response.render()` call;
  1. `my_action.html` to show the result of this action, after returning to the caller.

What is happening
-----

The right scenario (the first time I call the URL):

| `response.view` contents | when |
|-----------------------------------|---------|
| my_controller/my_view.html | `my_action` starts |
| div_layout.html | in `response.render` |
| my_controller/my_view.html | in `compileapp.run_view_in()`, after `my_action` |

The wrong scenario (in subsequent requests. Pay attention to the second line in table below):

| `response.view` contents | when |
|-----------------------------------|---------|
| my_controller/my_view.html | `my_action` starts |
| my_controller/my_view.html | in `response.render` |
| my_controller/my_view.html | in `compileapp.run_view_in()`, after `my_action` |

viniciusban:
Make `run_view_in()` use `response.view` from the environment, instead of `current.response`.

### DIFF
--- a/gluon/compileapp.py
+++ b/gluon/compileapp.py
@@ -652,7 +652,7 @@ def run_view_in(environment):
     """
     request = current.request
     response = current.response
-    view = response.view
+    view = environment['response'].view
     folder = request.folder
     path = pjoin(folder, 'compiled')
     badv = 'invalid view (%s)' % view


javierobcn:
I also had some problems with the views because of this. With this modification they have been resolved. Thanks

viniciusban:
Nice to know, Javier.

Let's see if the developers accept this PR.

javierobcn:
Web2py and community are fantastic. Thanks

PR #1080, opened 2015-10-06
Make ldap_auth a bit more DRY

BuhtigithuB:
Hello,

I make this part a little more DRY for further change that coming.

Thanks...

### DIFF
--- a/gluon/contrib/login_methods/ldap_auth.py
+++ b/gluon/contrib/login_methods/ldap_auth.py
@@ -419,32 +419,24 @@ def ldap_auth(server='ldap', port=None,
                     store_user_mail = result[user_mail_attrib][0]
                 except KeyError, e:
                     store_user_mail = None
-                try:
-                    #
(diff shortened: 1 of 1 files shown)


niphlod:
IMHO this should be handled by the dal layer.

BuhtigithuB:
I know, ldap_auth shouldn't update db, it makes no sens because it if
fields name change ldap_auth break... Though thing are how they are... To
me, ldap_auth should be rewrite... Each time I try to change something in
it, I get crazy...

:)

Richard

BuhtigithuB:
There is also a big issue with the way it handle username authentification
that cause in some circumstance the duplication of user record... This also
in part related to the fact that ldap_auth update directly the db without
passing by web2py... And to fix this, I had to change web2py and ldap_auth
because the later handle badly the username credentials...

mdipierro:
Yes, this should be handled by the DAL layer but, until then, this is an improvement.

BuhtigithuB:
:)

Issue #1095, opened 2015-10-29
Two-factor authentication improvement and bug fixed


josedesoto:
Working in the next patch:

When two-factor authentication is enables it looks that the last retry (last attempt to introduce the code) does not have effect. It did not have effect because the redirect and clean sessions is performed before to form.accepts. In the patch have been fixed.

I have done a few modification in tools.py to have two-factor more flexible, from models it is possible to do:

Enable two factor as global settings. It is compatible:
```
auth.setting.auth_two_factor_enabled = False
```

Set your custom attempts to login:
```
auth.setting.auth_two_factor_tries_left = 3
```

Message to return in case the code is incorrect:
```
auth.messages..invalid_two_factor_code = 'Incorrect code. {0} more attempt(s) remaining.'
```

To customize the email template:
```
auth.messages.retrieve_two_factor_code='Your temporary login code is {0}'
auth.messages.retrieve_two_factor_code_subject='Your temporary login code is {0}'
```

To customize two factor form:
```
auth.messages.label_two_factor = 'Authentication code'
auth.messages.two_factor_comment = 'This code was emailed to you and is required for login.'
```

It can be used to customize how to send the two factor code, or even to create your own factor code:
```
auth.settings.two_factor_methods = []
```

Once the two factor code is submit, you can use:
```
auth.settings.two_factor_onvalidation = []
```

By default if  ```auth.settings.two_factor_authentication_group``` is used or ```auth.setting.auth_two_factor_enabled``` is enable the code is generated as before and it is sent by email. The next examples are compatible with  two_factor_authentication_group or  ```auth_two_factor_enabled```. The only difference is that  auth_two_factor_enabled enable two-factor for the entire app, instead to use groups.

Example 1: We want to send the code by SMS, in our models:
```
def _sendsms(user, auth_two_factor):
    # we write the code to send the auth_two_factor by SMS
    return  auth_two_factor

auth.settings.auth_two_factor_tries_left = 2
auth.settings.auth_two_factor_enabled = True
auth.messages.two_factor_comment = "Your code have been sent by SMS"
auth.settings.two_factor_methods = [lambda user, auth_two_factor: _sendsms(user, auth_two_factor)]
```

Example 2: We want to send the code by SMS and create or own code:
```
def _sendsms(user, auth_two_factor):
    auth_two_factor = # we write the code to generate the code
    # we write the code to send the auth_two_factor by SMS
    return  auth_two_factor

auth.settings.auth_two_factor_tries_left = 2
[...]

PR #1096, opened 2015-10-29
issue #1095. Added two-factor authentication methods and onvalidation…

josedesoto:
When enable two-factor by group, the last attempt to retry to introduce the login does not have effect.  The reason is because the form.accepts is performed after to check the number of attempts left. 

The code as well has some modification to make two-factor more flexible, for more details:
https://github.com/web2py/web2py/issues/1095

If you are interested let me know and I can help to update the book with some example, as for example to integrate OTP with two-factor.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -1153,6 +1153,8 @@ class Auth(object):
         manager_actions={},
         auth_manager_role=None,
         two_factor_authentication_group = None,
+        auth_two_factor_enabled = False,
+        auth_two_factor_tries_left = 3,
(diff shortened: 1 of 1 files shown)


niphlod:
I'm positive that only a few are using it (including myself) exactly for the lack of documentation ^_^

The book is on github too, make a PR there https://github.com/mdipierro/web2py-book/

josedesoto:
PR done: https://github.com/mdipierro/web2py-book/pull/296

mdipierro:
Fantastic. Thank you!

PR #1098, opened 2015-11-05
Provide optional per-column SQLFORM.grid CSS class

BackSeat:
I wanted to control the CSS in the TD elements for one column in a SQLFORM.grid. I've added columnclasses={} to the grid() parameters, which is a dictionary of 'tablename.fieldname':'classname' e.g. {'t_mytable.f_myfield' : 'myclass'}. If you accept this pull request, I'll update the book pages (if you want: let me know).

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -1977,6 +1977,7 @@ class SQLFORM(FORM):
              user_signature=True,
              maxtextlengths={},
              maxtextlength=20,
+             columnclasses={},
              onvalidation=None,
              onfailure=None,
              oncreate=None,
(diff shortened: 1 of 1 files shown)


niphlod:
wouldn't a simple n-th css rule in your app's css be the appropriate place for such a thing ?

BackSeat:
I'm not sure I understand. Currently, the SQLFORM.grid columns don't have a CSS class; they are just <td>data</td>. My patch optionally makes that <td class="myclass">data</td>.

Are you saying there is already a way to apply a CSS class to, say, the fourth column (only) of a SQLFORM.grid?

niphlod:
yes, there is .
ref: http://www.w3schools.com/cssref/sel_nth-child.asp
example: http://codepen.io/srikarg/pen/BukHa

BackSeat:
Nice, I didn't know about that. Probably worth a paragraph in the SQLFORM.grid section of the book.

Issue #728, opened 2015-01-26
Failed login does not retain email/username field values (if form passes validation)


niphlod:
_From [markruol..._at_gmail.com](https://code.google.com/u/114619454780239522399/) on September 18, 2014 20:11:42_

What steps will reproduce the problem? 1. Login with auth.login(), with bad login info (but valid format to pass validation)
2. Login page will be redisplayed after login failure
3. Email/username field is empty What is the expected output? What do you see instead? The username/email field should retain its values after a failed login. Clearing the input forces the user to guess what they had previously typed upon login error, which is unnecessarily frustrating for the user. What version of the product are you using? On what operating system? Web2py 2.9.10 Please provide any additional information below. When auth.login() passes validation, but there is a login error, it redirects to it's own URL, where the username/email value is lost. https://groups.google.com/forum/#!topic/web2py/gXBfCGki9ns

_Original issue: http://code.google.com/p/web2py/issues/detail?id=1987_

PR #1109, opened 2015-11-27
Updated Auth.login() to remember input values after login attempt.

mrkrlli:
Updated Auth.login() to remember input values after failed login attempt wherre user is not found. This change does not remember passfield value for security reasons. Issue #728

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -2634,10 +2634,10 @@ class Auth(object):
                 accepted_form = False
 
                 if form.accepts(request, session if self.csrf_prevention else None,
-                                formname='login', dbio=False,
+                                formname='login', keepvalues=True, dbio=False,
                                 onvalidation=onvalidation,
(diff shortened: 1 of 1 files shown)


mrkrlli:
Setting the keepvalues argument to True does not keep the password input value when returning the form, which is good, but I'm exactly sure what causes this behavior.

mdipierro:
Please bring this up on the mailing list (or point me to it if you did already). this fix does not work becaue break existing correct behavior in other cases.

mrkrlli:
Will do, just posted this up on the mailing list: https://groups.google.com/forum/#!topic/web2py-developers/UGleCUF7DaY

mdipierro:
I will revisit this. Perhaps I dismissed it too soon.

abastardi:
Why does login() redirect after a failure rather than simply returning the login form, as proposed in pull request?

Issue #1123, opened 2015-12-09
Single attachment produces "'Int' object has no attribute lower" error


matclab:
As reported in 
https://groups.google.com/d/msg/web2py/kGuBn9iZByY/CJGFQYLWZ8oJ

Seen on GAE when sending a mail with
```
attachments = Mail.Attachment(…)
```

The error comes from gluon.py doing `enumerate(attachments)` on an object which is not iterable.

Pull request to come.

matclab:
closed with PR #1124.

PR #1124, opened 2015-12-09
Convert attachments to a list if necessary.

matclab:
Also corrects a typo that was apparently silenced by the bug.
This closes issue #1123


matclab:
Note that failing tests seem not related to the PR (Can't connect to MySQL server on 'localhost')

PR #1136, opened 2015-12-21
update files removing 2.5 things

cassiobotaro:
- Remove old python 2.5 things

### DIFF
--- a/web2py.py
+++ b/web2py.py
@@ -3,6 +3,9 @@
 
 import os
 import sys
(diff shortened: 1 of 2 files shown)


web2py.py
```diff
@@ -3,6 +3,9 @@
 
 import os
 import sys
+import gluon.widget
```
mdipierro: moving this here was the reason for 2.13.4

PR #1146, opened 2015-12-28
Jwt mcl

matclab:
Some improvement to jwt in web2py:
- Allows to encode date and datetime with json (needed when extra_fields of type date or datetime used for auth_user table)
- Allows jwt token to be produced with the logged in user …

This allows for jwt authentication without transmitting user and
password in clear. A use case is:

 - user first logs in,
 - he gets the token online over https,
 - he then issues API requests with the token as a HTTP header.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -117,6 +117,12 @@ def replace_id(url, form):
     return URL(url)
 
 
+def json_date_handler(obj):
+    if hasattr(obj, 'isoformat'):
+        return obj.isoformat()
+    else:
+        raise TypeError('%s is not JSON serializable' % obj)
+
 class Mail(object):
     """
     Class for configuring and sending emails with alternative text / html
@@ -1279,7 +1285,7 @@ class AuthJWT(object):
             if isinstance(secret, unicode):
                 secret = secret.encode('ascii', 'ignore')
         b64h = self.cached_b64h
-        b64p = self.jwt_b64e(json_parser.dumps(payload))
+        b64p = self.jwt_b64e(json_parser.dumps(payload, default=json_date_handler))
         jbody = b64h + '.' + b64p
         mauth = hmac.new(key=secret, msg=jbody, digestmod=self.digestmod)
         jsign = self.jwt_b64e(mauth.digest())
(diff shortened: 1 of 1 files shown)


matclab:
Deserialization of date and datetime is not taken into account.
We should probably use gluon.serialize somewhere…

gluon/tools.py
```diff
@@ -1393,16 +1400,16 @@ def api_auth():
...
             if valid_user:
                 payload = self.serialize_auth_session(current.session.auth)
                 self.alter_payload(payload)
                 ret = {'token':self.generate_token(payload)}
             else:
                 raise HTTP(
-                    401, u'Not Authorized', 
+                    401, u'Not Authorized',
                     **{'WWW-Authenticate': u'JWT realm="%s"' % self.realm})
-        else:
```
mdipierro: why is this removed? shouldn't we handle this case?

matclab: It seems I've messed the PR. There is some indentation mess.
The idea was to keep only the 'Not Authorized' redirection…
I'll have a second take to this PR… after a good night.

matclab:
@mdipierro what is the *web2py* way to cast a string back to a field type ? 
I'd like to do the following in `inject_token`:
```python
     self.auth.user = Storage(tokend['user'])
     for field in self.auth.settings.extra_fields[self.auth.settings.table_user_name]:
         self.auth.user[field.name] = cast(field.type, self.auth.user[field.name])
```
But I do not know how to do the `cast`.
Is this transformation already done sowewhere in web2py ?

mdipierro:
You cannot really cast that way because the conversion may not be possible. I would not do the casting.

matclab:
@mdipierro I've taken into account your comment.

The new commits, only make use of gluon.serializers in order to take into account more types when serializing to json, and allow to generate the token in the browser when connected. 
I've also corrected a typo and added a comment about using before_authorization callback for extra fields…

niphlod:
thanks for the PR. I needed to make further modifications to it so I raised https://github.com/web2py/web2py/pull/1149 that contains this and another fix.

matclab:
Thanks !

PR #1150, opened 2015-12-30
better test_utils and reorg to compileapp

niphlod:


niphlod:
will do another PR, sorry.

PR #964, opened 2015-05-09
#963 : Convert subject and body to unicode before sending mail

matclab:
Patch for issue #963.

Not doing this was raising an exception :

Traceback (most recent call last):
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/gluon/restricted.py, line 227, in restricted
    exec ccode in environment
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/applications/foundit/controllers/default.py, line 127, in <module>
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/gluon/globals.py, line 393, in <lambda>
    self._caller = lambda f: f()
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/applications/foundit/controllers/default.py, line 34, in user
    return dict(form=auth())
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/gluon/tools.py, line 1595, in __call__
    return getattr(self, args[0])()
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/gluon/tools.py, line 3272, in request_reset_password
    if self.email_reset_password(user):
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/gluon/tools.py, line 3296, in email_reset_password
    message=self.messages.reset_password % d):
  File /base/data/home/apps/e~sacred-bonus-88417/1.384178859090314065/gluon/tools.py, line 798, in send
    subject=subject, body=text, **xcc)
  File /base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/api/mail.py, line 402, in send_mail
    message.send(make_sync_call=make_sync_call)
  File /base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/api/mail.py, line 1108, in send
    message = self.ToProto()
  File /base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/api/mail.py, line 1350, in ToProto
    message = super(EmailMessage, self).ToProto()
  File /base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/api/mail.py, line 1046, in ToProto
    message.set_subject(_to_str(self.subject))
  File cpp_message.pyx, line 124, in cpp_message.SetScalarAccessors.Setter (third_party/apphosting/python/protobuf/proto1/cpp_message.cc:2229)
[...]

PR #1163, opened 2016-01-07
Fix bug on Mail.send() when text or input are Unicode

rafaelol:
On PR #964 @matclab forced the encoding of both subject and
text variables to unicode.

After merging it, matclab realized that when we send Unicode
text to the method it raises an exception and asked if he should
change the commit. Unfortunately this thing was kept untouched.

This problem exists because we previously encode the unicode variables
to utf-8 (for instance here https://github.com/web2py/web2py/blob/master/gluon/tools.py#L478-L481) 
and then convert to unicode again. This piece of code shows the exception being raised:

```
>>> a = u'áéí'
>>> a
u'\xe1\xe9\xed'
>>> b = a.encode('utf-8')
>>> b
'\xc3\xa1\xc3\xa9\xc3\xad'
>>> unicode(a)
u'\xe1\xe9\xed'
>>> unicode(b)
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
```

If we force to str, just like @matclab suggested, we solve this issue.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -777,16 +777,16 @@ class Mail(object):
                 if attachments:
                     result = mail.send_mail(
                         sender=sender, to=origTo,
-                        subject=str(subject), body=str(text), html=html,
+                        subject=unicode(subject, encoding), body=unicode(text, encoding), html=html,
                         attachments=attachments, **xcc)
                 elif html and (not raw):
                     result = mail.send_mail(
                         sender=sender, to=origTo,
-                        subject=str(subject), body=str(text), html=html, **xcc)
+                        subject=unicode(subject, encoding), body=unicode(text, encoding), html=html, **xcc)
                 else:
                     result = mail.send_mail(
                         sender=sender, to=origTo,
-                        subject=str(subject), body=str(text), **xcc)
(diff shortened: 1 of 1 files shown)


gluon/tools.py
```diff
@@ -777,16 +777,16 @@ def encoded_or_raw(text):
                 if attachments:
                     result = mail.send_mail(
                         sender=sender, to=origTo,
-                        subject=unicode(subject), body=unicode(text), html=html,
+                        subject=str(subject), body=str(text), html=html,
```
cassiobotaro: Maybe a better solution is 
```diff
- subject=str(subject), body=str(text), html=html,
+ subject=unicode(subject, encoding), body=unicode(text, encoding), html=html,
```

rafaelol: Great! I think it's better than matclab's solution.

Will send another commit to the PR.

matclab: Yeap, I've come to the same solution on my local tree, but totally forgot
to submit a new PR…
I'm sorry for that !

Thanks a lot Raphael for reviving it !

-Mathieu

2016-01-07 17:57 GMT+01:00 Rafael Oliveira Lopes <notifications@github.com>:

PR #1166, opened 2016-01-11
Make SessionSetDb.get() a generator.

Rimbo:
Make the same change for SessionSetDb that's in SessionSetFiles: Make get() a generator, so that memory doesn't get blown out if there are 18 million sessions there to clean up.


mdipierro:
nice

PR #1193, opened 2016-02-21
Fixed a bug whereby calling 'delete_record()' on a row object caused …

CzechErface:
…an AttributeError.

The error is reproducible by creating an SQLFORM and specifying the 'deletable' parameter as a callback.  Upon deletion Web2Py will encounter an AttributeError because there is no such attribute of 'delete_record' for the 'record' object.  I simply copied the behavior of the block  where deletable is NOT callable by overwriting a single line of code.  This could be simplified, but I am not experienced enough with the DAL to do such a thing.


CzechErface:
I am new to open source contributions.  What can I do to increase the
coverage of my pull request?

PR #1195, opened 2016-02-23
fixes #1191

### LINKED ISSUE #1191: Scheduler - Prevent Drift and Failed Runs - Next Run Time not progressing.
Please refer to this thread over on the main web2py usergroup. https://groups.google.com/forum/#!topic/web2py/khykEwJJrmY

I believe that there is a bug in gluon/scheduler.py which is causing tasks that utilize the Prevent Drift option to run repeatedly when a run fails. Prevent Drift utilizes a times_run variable when calculating when the next run should occur, however when tasks do not complete successfully this count does not currently get incremented which means that the next run time doesn't get moved forward and thus the task just gets picked up again almost immediately.  Review of the relevant source code suggests to me that the error is occurring in the section between lines 911 and 932 and that a new line should be inserted between lines 930 & 931 to update the times_run value as is done on line 914 for successful runs.

```
#gluon/scheduler.py starting at line 911
        if task_report.status == COMPLETED:
            d = dict(status=status,
                     next_run_time=task.next_run_time,
                     times_run=task.times_run,
                     times_failed=0
                     )
            db(st.id == task.task_id).update(**d)
            if status == COMPLETED:
                self.update_dependencies(db, task.task_id)
        else:
            st_mapping = {'FAILED': 'FAILED',
                          'TIMEOUT': 'TIMEOUT',
                          'STOPPED': 'FAILED'}[task_report.status]
            status = (task.retry_failed
                      and task.times_failed < task.retry_failed
                      and QUEUED or task.retry_failed == -1
                      and QUEUED or st_mapping)
            db(st.id == task.task_id).update(
                times_failed=db.scheduler_task.times_failed + 1,
                next_run_time=task.next_run_time,
                times_run=task.times_run, #<-- Insert this line to fix
                status=status
            )

```
Please review this suggested bug fix and consider for inclusion.

niphlod:
fix is in lines 828-835 . needed to backport total_seconds for py2.6
(694-701).
everything else is just pep8.

### DIFF
--- a/gluon/scheduler.py
+++ b/gluon/scheduler.py
@@ -215,7 +215,7 @@ class JobGraph(object):
                 nested_dict = dict(
                     (item, (dep - ordered)) for item, dep in nested_dict.items()
                     if item not in ordered
-                    )
+                )
             assert not nested_dict, "A cyclic dependency exists amongst %r" % nested_dict
             db.commit()
             return rtn
@@ -297,7 +297,7 @@ def executor(queue, task, out):
             f = task.function
             functions = current._scheduler.tasks
             if not functions:
-                #look into env
+                # look into env
                 _function = _env.get(f)
             else:
                 _function = functions.get(f)
@@ -314,7 +314,7 @@ def executor(queue, task, out):
             vars = loads(task.vars, object_hook=_decode_dict)
             result = dumps(_function(*args, **vars))
         else:
-            ### for testing purpose only
+            # for testing purpose only
             result = eval(task.function)(
                 *loads(task.args, object_hook=_decode_dict),
                 **loads(task.vars, object_hook=_decode_dict))
@@ -663,7 +663,7 @@ class Scheduler(MetaScheduler):
             Field('traceback', 'text'),
             Field('worker_name', default=self.worker_name),
             migrate=self.__get_migrate('scheduler_run', migrate)
-            )
+        )
 
         db.define_table(
             'scheduler_worker',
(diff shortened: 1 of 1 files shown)


gluon/scheduler.py
```diff
@@ -817,11 +825,14 @@ def pop_task(self, db):
...
-        else:
-            next_run_time = task.start_time + datetime.timedelta(
-                seconds=task.period * times_run
             )
+        else:
+            # calc next_run_time based on available slots
+            # see #1191
+            secondspassed = self.total_seconds(now - task.start_time)
+            steps = secondspassed // task.period + 1
+            next_run_time += datetime.timedelta(seconds=task.period * steps)
```
bmeredyk: Nope. next_run_time hasn't even been used before so this will always cause an exception.

Should be this

`next_run_time = task.start_time + datetime.timedelta(seconds=task.period * steps)`

gluon/scheduler.py
```diff
@@ -912,7 +923,7 @@ def report_task(self, task, task_report):
             d = dict(status=status,
                      next_run_time=task.next_run_time,
                      times_run=task.times_run,
-                     times_failed=0
+                     # times_failed=0
```
bmeredyk: I don't think you really want this commented out. If times_failed doesn't get reset after each successful run then it becomes a cumulative failure counter over the task's entire lifetime so over time even very rare failures could cause the times_failed to surpass the allotted retries and the task to incorrectly land in the FAILED state.

bmeredyk:
Not quite ready for prime time. Two issues:

line 834 will always fail because next_run_time doesn't exist yet. should be 
```next_run_time = task.start_time + datetime.timedelta(seconds=task.period * steps)```

line 926 probably shouldn't be commented out else it is a cumulative failure counter and not just a count of failures since the last success.

niphlod:
okey. first issue is a huge mistake on my side, I removed too many lines before committing :°°D

second issue: I'd leave it there or not but looking closely it's another "misbehaviour" ? if a task is set to repeat until 5 failures happen (e.g. because it's a call to an API that costs you money every time you hit it, and you can't afford too many failures) why times_failed should be reset as soon as a task gets completed ?

bmeredyk:
Sure, I see your point on the failure count and retries but it also represents a big change from the behavior of previous versions. Changing the behavior now, even if documented in the book and announced on the user group, just seems like a change that is waiting to catch users off guard when a task suddenly goes FAILED even though it seems to have been running fine other than rare errors that went unnoticed.
I think given the difference from historic behavior if the change is implemented there should also be a new option to pick between the old behavior (default) and your proposed new one. A new boolean field for the tasks take and then a quick check of its value before the counter get updated after successful runs.

niphlod:
ok, will revert that bit. Then we can talk about new features to implement

niphlod:
voilà.

PR #1201, opened 2016-03-04
Czech translations (admin,welcome) + small script for translations + nginx/uwsgi/ubuntu script

zvolsky:
Hi Massimo and others.
I am sorry, I am not able split the push request into smaller pieces.

So, what is here:

1. newer cs.py in admin and welcome.

2. script to set untranslated messages in translation file from another one (if you want update application file from newer version of welcome file, if you start translation for very similar language, and so on).

3. nginx/uwsgi/ubuntu script for systemd (creates /etc/systemd/system/emperor.uwsgi.service as described here: https://groups.google.com/forum/#!topic/web2py/ZM9IIEjtHSI + uses: service..start/service..stop)
This worked good for me at forpsi virtual debian 8.
However I am not good in shell scripting. So some review and testing here would be good. But no idea how to do it.
Especially mkdir commands: what happens if directories already exist. I think it generates no fatal failure, but warnings.

No problem if something or all is acceptable. It is just first attempt with pull request here.
If later I will sure create separate pieces.
Thank you.
Mirek Zvolsky


mdipierro:
Can you please resend this patch but leave "uwsgi-emperor" inchanged?

zvolsky:
this is replaced with 3 separate pull requests, 6 days later -- remove this

PR #1209, opened 2016-03-10
new script to update untranslated messages from other language file

zvolsky:


mdipierro:
For now I will take this, there is a general consensus we want to reduce the content of scripts, not include it.

PR #1219, opened 2016-03-15
Improve PEP8

BuhtigithuB:


gluon/html.py
```diff
@@ -2587,16 +2619,16 @@ def xml(self):
...
+def embed64(filename=None,
+            file=None,
+            data=None,
+            extension='image/gif'
+            ):
     """
     helper to encode the provided (binary) data into base64.
 
     Args:
+        extension:
```
niphlod: If you're making a point on rearraging the code this should come AFTER every other argument :P

gluon/html.py
```diff
@@ -2339,6 +2366,10 @@ def as_dict(self, flat=False, sanitize=True):
 
         Sanitize is naive. It should catch any unsafe value
         for client retrieval.
+
+        Args:
+            sanitize:
```
niphlod: args in reverse order....

gluon/html.py
```diff
@@ -437,9 +429,20 @@ def verifyURL(request, hmac_key=None, hash_vars=True, salt=None, user_signature=
         >>> verifyURL(r, 'key')
         False
 
+    Args:
+        user_signature:
```
niphlod: args in wrong order, too

BuhtigithuB: I just append args automatically with pycharm sorry if they are not in the
proper order... Can be fixed latter...

gluon/html.py
```diff
@@ -186,6 +188,8 @@ def URL(
     within the present application and controller.
 
     Args:
+        language:
```
niphlod: here too . Another point for brains over IDEs with fancy plugins....

gluon/html.py
BuhtigithuB: You want them in the same order of the method signature? I think pycharm
put them in alphabetical order


On Tue, Mar 15, 2016 at 4:37 PM, Richard Vézina <ml.richard.vezina@gmail.com
> wrote:

> I just append args automatically with pycharm sorry if they are not in the
> proper order... Can be fixed latter...
>

gluon/html.py
BuhtigithuB: Niphold please state what you want and I fix that...

niphlod:
I'd drop and recreate it with corrections needed...

BuhtigithuB:
but what is the common issue denominator? order of args in docstring?

niphlod:
those docstrings go to API docs automatically built. No docstrings are bad, wrong docstrings are worse....pycharm is an IDE, not a python developer ^_^

BuhtigithuB:
so you want I remove entirely the append you specified ? that it?

niphlod:
it's not the first time it pops up: pep8 is good and all, but not everybody sees red warnings because not everybody uses pycharm (or, specifically, your settings for pycharm). want to rearrange the code according to pep8 specs? ok, usually it's more readable so it's usually a good contribution.
Don't be an autopep8 guy, though... put some thoughts in it
If you're going to rewrite docstrings, please watch at the bigger picture: docstrings are for documenting the API (yes, also in IDEs). Meaning a guy can skip over reading the code and get a humanly description on what the method does an what arguments it accepts.
In web2py's case, they even get published over at readthedocs.org 
Adding empty arguments without even blinking, and in the wrong order ... ahem ... it helps nobody. It actually goes against "proper documentation wishes".

BuhtigithuB:
Agree with that... About the rest of PEP8 they are made manually so no autopep8... I just append args as it shut up warning of pycharm... I thought it would be better to include the args then leave them not include... In the future I will avoid touching docstring except if I find the time to write them properly.

Helping me learn would be a better way to keep me helping with what I undertake instead of that I feel that you are angry because of this contrib which to me except some of the things you spotted was pretty good...

niphlod:
don't get me wrong but you shouldn't feel anything if you get comments on your PRs . For all matter and purposed, nobody rejected it. IMHO though you didn't get the point as https://github.com/web2py/web2py/pull/1220 brings the exact same issues.

BuhtigithuB:
Should fix the problem : https://github.com/web2py/web2py/pull/1224

PR #1220, opened 2016-03-15
Improve PEP8

BuhtigithuB:


niphlod:
see https://github.com/web2py/web2py/pull/1219#issuecomment-197013653

PR #1230, opened 2016-03-19
test_IS_IMAGE and more

leonelcamara:
* test for IS_IMAGE validator
* test for IS_IN_SET options
* Removed unused and undocumented IS_IN_SUBSET
* Fixed _options in IS_EMPTY_OR not passing arguments to the other options method
* Some tests for formatters


leonelcamara:
This increases coverage in validators.py from last week's 78.21% to 82.34% right now **all** the validators are at least partially tested.

BuhtigithuB:
You are rigth to say partially as I had look at coverage and if understand
it properly there is many validators "if" conditions which are not
tested... I would say that this is the next step : analysis the existing
test case and make sure that every code line of them are covered as much as
possible...

:)

PR #1235, opened 2016-03-22
Avoid issuing a separate sql query for each row in sqlform grid search results, for virtual fields

kjkuan:
When deriving the value for a virtual field in a sqlform grid, this change avoids issuing a separate sql query for each row found in the search results.

Feel free to ignore the PR or simply see it as a bug report if my change doesn't make sense.

Thanks

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -2743,7 +2743,11 @@ class SQLFORM(FORM):
                     if field.type == 'blob':
                         continue
                     if isinstance(field, Field.Virtual) and field.tablename in row:
-                        value = row[field.tablename][field.name]
+                        try:
+                            # fast path, works for joins
+                            value = row[field.tablename][field.name]
+                        except KeyError:
+                            value = dbset.db[field.tablename][row[field.tablename][field_id]][field.name]
                     else:
                         value = row[str(field)]
                     maxlength = maxtextlengths.get(str(field), maxtextlength)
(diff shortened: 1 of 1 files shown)


mdipierro:
Sorry. you change will work in some cases but not all. The grid row may not contain all the info it needs to compute the virtual field.

kjkuan:
okay, but is it not worth it to consider the case in which the row does have the info it needs and then use it directly instead of always issue a separate SQL query?

Or is it something that can be done in pydal? Could you help me understand the case in which the change proposed by the PR would fail? There must be something we could do, right? If not, please help me understand... Thanks

mdipierro:
Let me think about this and respond later today. I agree there are some case when this would work.

kjkuan:
Thank you. I'd reopen this PR if I could, but I can't.

Perhaps the special case would be to check additionally if `field.name in row[field.tablename]`? And if so, then `value = row[field.tablename][field.name]` otherwise do `value = ...` assignment that generates the SQL query as before. Obviously, I'm not familiar with PyDAL..., just a guess.

mdipierro:
I reopened but it needs a little but of work. I can do it.

mdipierro:
How about a try(your syntax) execpt (the old syntax)?

mdipierro:
Can you update the PR so you get credit for this?

kjkuan:
thank you, updated the PR.

PR #1238, opened 2016-03-23
BEAUTIFY, MENU, MARKMIN and ASSIGNJS basic tests and attempt ASSIGNJS…

BuhtigithuB:
… documentation

### DIFF
--- a/gluon/html.py
+++ b/gluon/html.py
@@ -2831,10 +2831,24 @@ class MARKMIN(XmlComponent):
 
 
 def ASSIGNJS(**kargs):
+    """
+    Example:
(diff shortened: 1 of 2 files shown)


BuhtigithuB:
Grrr... Some asserts should hit some more lines but doesn't seems to work...

:)

mdipierro:
There are lots of markmin tests in markmin.py itself. we should perhaps delegate. Anyway this is good and will go in. Thank you!

BuhtigithuB:
I think the entire MARKMIN() is cover... I did it for raise coverage, I
don't think it a problem to have some duplicate, they are not really
duplicate anyway, as long as we don't start test markmin.py...

PR #1246, opened 2016-03-24
Improve PEP8 just a bit + more coverage

BuhtigithuB:


BuhtigithuB:
We may reach 42% one day... 

:D

PR #1252, opened 2016-03-26
Improve test chache

BuhtigithuB:


BuhtigithuB:
Tests will follow when I have more time...

PR #1253, opened 2016-03-27
More test coverage for validators.py

leonelcamara:
Fixed a bug in IS_EMAIL throwing exceptions when asked to validate anything other than a string which was problematic for ANY_OF
Fixed a bug in ANY_OF.formatter where it was trying to format with a validator that didn't validate


BuhtigithuB:
We make it to 42%

:D

PR #1254, opened 2016-03-30
Seriously increase template.py coverage

leonelcamara:
test block extend, super, delimiters, etc.


BuhtigithuB:
Nice! Keep going...

mdipierro:
Fantastic!

PR #1255, opened 2016-03-30
enable translation for tools.Expose

chenl:
Add T(...) for user facing strings: "Folders" and "Files"


mdipierro:
I do not think this works. Have you tested it?

chenl:
This will hopefully teach me to not send pull requests w/o testing code.

PR #1260, opened 2016-04-02
SQLFORM() basic check

BuhtigithuB:


BuhtigithuB:
Don't know why codecov complains... Maybe to many characters in the test case??

BuhtigithuB:
Maybe related to that : 

https://github.com/codecov/support/issues/135

Not sure why the coverage had drop as much.

BuhtigithuB:
Maybe since there were no test_sqlhtml.py the coverage calculation were computed differently and it failed because coverage lower so much since the actual coverage provide by test_sqlhtml.py is low... Though the actual test don't seems to reach any of their target reading the report from codecov...

mdipierro:
This test scares me a little. It checks for the exact HTML structure of the form. Adding an attribute to a field (for example a placeholder attribute or no autocomplete) would break the test even if it is an improvement of the code. At the same time it is nice to know that we have changed something that changes the html of the form. I will think about this some more.

BuhtigithuB:
It up to you... I know they are really dumb, but I considering that sqlhtml
will shade away... Having them allow us to know that a commit over
sqlhtml.py didn't break anything and at the same time incrise the coverage
which is my main goal... I know that wiser tests would be better and you
may consider not merging the PR... I will try to improve them later and
send another PR...

PR #1262, opened 2016-04-04
initial tests for scheduler.py, plus a minor fix for JobGraph

niphlod:
of course this is a stub. 
Apparently extricating coverage subtleties with subprocess is impossible. Any help to implement it is greatly appreciated (or else nothing called by "executor", albeit run, won't be reported as covered).
That being said, unittesting the behaviour of the scheduler is now possible. 
check TestsForSchedulerRunner which can be used as a testbed to implement further tests. 
The format is a decent copy/paste from w2p_scheduler_tests, which is ATM the de-facto suite for the scheduler.
This adds ~90 secs to the duration of the testsuite at each and every testclass. Theoretically we should avoid having multiple "*Runner" tests because at each time the env is reset and a new scheduler gets launched.
The scheduler is configured to die after 10 empty loops and albeit the heartbeat has been reduced to 1 sec, it still needs quite some time.

Of course if someone has a better alternative it's greatly appreciated too ^_^

Happy iterations.

### DIFF
--- a/gluon/scheduler.py
+++ b/gluon/scheduler.py
@@ -181,7 +181,7 @@ class JobGraph(object):
                                            task_child=task_child,
                                            job_name=self.job_name)
 
-    def validate(self, job_name):
+    def validate(self, job_name=None):
         """Validates if all tasks job_name can be completed, i.e. there
         are no mutual dependencies among tasks.
         Commits at the end if successfull, or it rollbacks the entire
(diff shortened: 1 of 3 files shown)


niphlod:
umpf. travis on my account didn't fail .......

BuhtigithuB:
may pytest provide more flexibility to reused (share) elements that impact performance of the test? I found unittest more limited in this aspect...

niphlod:
not that I'm aware of but you're welcome to try and show your accomplishments.

PR #1263, opened 2016-04-04
initial tests for scheduler.py, plus a minor fix for JobGraph

niphlod:
...and apparently nobody is running the test-suite at home ;-( dal tests required you to have a fully configured travis-like env with mysql AND postgresql available .... fixed that too.

Back on topic, this is a stub. 
Apparently extricating coverage subtleties with subprocess(es) is impossible. Any help to implement it is greatly appreciated (or else nothing called by "executor", albeit run, won't be reported as covered).
That being said, unittesting the behaviour of the scheduler is now possible. 
check TestsForSchedulerRunner which can be used as a testbed to implement further tests. 
The format is a decent copy/paste from w2p_scheduler_tests, which is ATM the de-facto suite for the scheduler.
This adds ~90 secs to the duration of the testsuite at each and every testclass. Theoretically we should avoid having multiple "*Runner" tests because at each time the env is reset and a new scheduler gets launched.
The scheduler is configured to die after 20 empty loops and albeit the heartbeat has been reduced to 1 sec, it still needs quite some time.

Of course if someone has a better alternative it's greatly appreciated too ^_^

Happy iterations.

### DIFF
--- a/gluon/scheduler.py
+++ b/gluon/scheduler.py
@@ -181,7 +181,7 @@ class JobGraph(object):
                                            task_child=task_child,
                                            job_name=self.job_name)
 
-    def validate(self, job_name):
+    def validate(self, job_name=None):
         """Validates if all tasks job_name can be completed, i.e. there
         are no mutual dependencies among tasks.
         Commits at the end if successfull, or it rollbacks the entire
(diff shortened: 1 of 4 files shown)


leonelcamara:
> ...and apparently nobody is running the test-suite at home ;-( dal tests required you to have a fully configured travis-like env with mysql AND postgresql available .... fixed that too.

I'm running it, I was just ignoring the dal tests because of how annoying that was. Thanks!

PR #1269, opened 2016-04-06
Test Mail.Attachment by mailing test_tools.py itself

leonelcamara:


BuhtigithuB:
watchout merge hell!!

leonelcamara:
@BuhtigithuB it should be ok since I only touched the Mail test in test_tools.py and you only changed Auth, it should be able to merge automatically.

@niphlod I got the scheduler test to count subprocesses in coverage by running the scheduler using coverage itself, make sure to review that commit

niphlod:
@leonelcamara : uhm ... sure that it works for you? Before pushing my PR I tried pretty much every variation already and got some times similar errors as current tests.....both travis-ci and appveyor are saying it's not working, and my computer too ... does it really work on yours ?

leonelcamara:
@niphlod It works for me, it doesn't work on travis-ci, I think it's because it's using coverage 4.0.3 while I'm using 4.1b2 is it possible to tell travis-ci to use a newer version of coverage?

niphlod:
doesn't even on appveyor though......... let me try on my computer.

leonelcamara:
appveyor is also using 4.0.3

niphlod:
well, instead of make tests fail, given that 4.1b2 is a beta, I'd stay with 4.0.3 until it gets released... even if that means not having code coverage for 40 lines of scheduler.py

leonelcamara:
Agreed, I'll remove that commit I guess I'll try again when coverage 4.1 is properly released.

PR #1271, opened 2016-04-06
delete unused reserved_sql_keywords.py because the DAL uses this one …

leonelcamara:
…pydal/contrib/reserved_sql_keywords.py


leonelcamara:
@codecov-io learn to count 45.88% is 0.04% increased coverage compared to 45.84%

BuhtigithuB:
:)

PR #1272, opened 2016-04-07
New logout_bare() for shell logout and refactor test using it

BuhtigithuB:
Not exactly sure it fully correct, but tests pass...


BuhtigithuB:
It learn to count!!

PR #1282, opened 2016-04-09
Feature/heading anchor

BuhtigithuB:


mdipierro:
I am find with all the pep8 improvements but do we really have to add this extra flag? Can we always have default heading-anchors or handle this via JS in the books app?

BuhtigithuB:
About defaulting it, I don't know... I don't know what use markmin2html.py someone may not want the extrat div and link if they don't use the feature... It surely feasible with JS, though I am a JS dumb...

BuhtigithuB:
JS : http://renaysha.me/anchorific-js/

BuhtigithuB:
It not working exactly as what I did as in the actual implementation the anchor is invisible shows up when mouse over the line (not need to over the heading or the anchor it self), but it may be configurable... Didn't check...

mdipierro:
yes or simply something like this:

jQuery('h1,h2,h3').not('[id]').each(function(){var t=jQuery(this);t.attr('id',t.text().replace(/[^ \w-]/ig,'').toLowerCase().replace('-'));});

BuhtigithuB:
Massimo the jQuery that you sent seems to be cropped... I take a look at markmin2html and I don't think I can make this work without a flag... So if it works with jQuery let go for it.

mdipierro:
Can you remove the anchor logic from this PR? If you do I can fix the book app with the jQuery.

BuhtigithuB:
Done!

BuhtigithuB:
NOTE : The CSS in web2py-book app is pretty clean headings should be like this : 
<div class="hover-show-headerlink"><hX><a class="hearderlink" href="#title-name"</hX></div>

PR #1284, opened 2016-04-10
More markmin pep8

BuhtigithuB:


BuhtigithuB:
Once there here the following PEP8 for markmin contrib...

PR #1290, opened 2016-04-12
box-sizing best practice

carpaIdea:
This code was updated to match new box-sizing best practices. Also prefixes are pretty much dead. (quote from http://www.paulirish.com/2012/box-sizing-border-box-ftw/).

More info on https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/

### DIFF
--- a/applications/examples/static/css/stupid.css
+++ b/applications/examples/static/css/stupid.css
@@ -5,7 +5,8 @@
 ************/
 
 /*** basic styles ***/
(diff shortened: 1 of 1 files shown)


mdipierro:
Thanks for all these fixes. Notice stupid.css is its own github project so I am copying your patches there too. Please submit your patches there next time.

PR #1291, opened 2016-04-13
Test emails with alternative text and html

leonelcamara:


mdipierro:
Thanks. This helps a lot.

PR #1292, opened 2016-04-13
Some more coverage for XML() and DIV()

BuhtigithuB:


BuhtigithuB:
@leonelcamara, could it be caused by the change you made about pypy?

leonelcamara:
@BuhtigithuB sadly no.

Appveyor couldn't recursively clone the DAL repository (that's a new one).

As for travis it was caused by the scheduler tests subprocess.call hanging again, I hoped using a more modern pypy version would solve that problem but apparently it can still happen... FML. Hopefully it will at least make the issue less recurrent.  
  
This issue with pypy and the scheduler tests is very hard to debug because of how random it is.

leonelcamara:
Actually this seems a bit different. It said error popping tasks which may indicate it couldn't connect to the scheduler database.

PR #1295, opened 2016-04-14
Refactor Auth tests, new test, old implementation commented for now

BuhtigithuB:
Test of login_bare() is failing for now as login_bare() always return false, I think it is broken, see here : 
https://groups.google.com/forum/#!topic/web2py-developers/z1VgPNVgzLA


BuhtigithuB:
Make it 47%...

BuhtigithuB:
Scheduler!!

@leonelcamara @niphlod why it seems to be more often with 2.7?

PR #1296, opened 2016-04-14
Enhancement tools.py PEP8

BuhtigithuB:


gluon/tools.py
```diff
@@ -2950,7 +2945,7 @@ def login(self,
                                )
 
                 captcha = settings.login_captcha or \
-                    (settings.login_captcha != False and settings.captcha)
+                    (settings.login_captcha is notFalse and settings.captcha)
```
cassiobotaro: i think you miss a space here

BuhtigithuB: yes good catch...

BuhtigithuB:
Fixed...

gluon/tools.py
BuhtigithuB: Fixed

BuhtigithuB: Now fixed... Forget to git add tools.py :(

PR #1298, opened 2016-04-16
New test suite for prettydate() + fix wrong number of days for month

BuhtigithuB:
More tests!!

### DIFF
--- a/gluon/tests/test_tools.py
+++ b/gluon/tests/test_tools.py
@@ -7,6 +7,7 @@
 import os
 import sys
 import smtplib
+import datetime
 if sys.version < "2.7":
     import unittest2 as unittest
 else:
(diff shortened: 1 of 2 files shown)


BuhtigithuB:
I think this : 
```python
-    elif dt.days > 21:
+    elif dt.days >= 30:
```
was a bug...

??

leonelcamara:
The days >21 was not really a bug more of a rounding up. I don't mind changing it but I wouldn't pick 30 because otherwise you will get stuff like 4 weeks which just looks bad because it would look better as 1 month.  So if you're going to change that change it to something between 21 and 27.

BuhtigithuB:
@leonelcamara good point... and thinking to that with > 21 you got 3 weeks, I was concerned about that.. I will revert it...

BuhtigithuB:
doh! you are rigth with 27...

BuhtigithuB:
* [x] prettydate covered

leonelcamara:
Ahahahah I like the comment, and it was needed so it doesn't cause confusion again in the future.

BuhtigithuB:
haha!

BuhtigithuB:
Tests pass previously here, the only change I made was fixing a typo in a comment "if -> is"... 

Can be merge.

PR #1305, opened 2016-04-21
100% coverage for recfile

leonelcamara:


gluon/tests/test_recfile.py
```diff
@@ -24,7 +24,7 @@ def tearDown(self):
         shutil.rmtree('tests')
 
     def testgeneration(self):
```
BuhtigithuB: Is this a test_ function? If so can you fix test name... Thanks...

leonelcamara: Sure

Issue #1310, opened 2016-04-27
compiled app's views aren't executed


niphlod:
This PR https://github.com/web2py/web2py/pull/1194 broke compiled apps. 
Before reverting and releasing, I propose a reassessment of the whole compileapp.py (seems that we can order both code and "thoughs" once and for all). I opened a relevant discussion on web2py-developers

PR #1311, opened 2016-04-28
Bug fix request

fpardourrutia:
I am developing a Mac Desktop application using Web2py. My application
worked perfect before compilation. However, after compilation, my
custom views could not be displayed, so the generic ones were used.

I figured out that the names of the compiled views were not consistent
with the ones of the other compiled files. For example:
views_01_conglomerado_index.html.pyc

instead of:
views.01_conglomerado.index.html.pyc

I have not tried in Windows, but maybe the correction should be done as
well, yielding:

filename = ('views/%s.py' % fname).replace('/', '.').replace('\\', '.')

### DIFF
--- a/gluon/compileapp.py
+++ b/gluon/compileapp.py
@@ -480,7 +480,7 @@ def compile_views(folder, skip_failed_views=False):
             else:
                 raise Exception("%s in %s" % (e, fname))
         else:
-            filename = ('views/%s.py' % fname).replace('/', '_').replace('\\', '_')
+            filename = ('views/%s.py' % fname).replace('/', '.').replace('\\', '_')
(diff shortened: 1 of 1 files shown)


fpardourrutia:
Test case:
https://github.com/tereom/cliente_snmb/commit/45fc2a5b8ccd87fde5d0450563a9ac21858f2c86

BuhtigithuB:
Think it a known problem :
https://groups.google.com/d/msg/web2py-developers/4a1izI4W8PI/5f4ZmgETAQAJ

niphlod:
tracked at https://github.com/web2py/web2py/issues/1310 . Will close this as we (devs) are thinking to simplify all the surrounding code in addition to fix the issue. Thank for the report though

PR #1312, opened 2016-04-30
security issue#1261: gluon.tools.Expose symlinks

chenl:
gluon.tools.Expose will follow symlinks outside of base directory.

By default don't allow to follow symbolic links outside of the exposed
directory, unless follow_symlink_out is set to True in
Expose.__init__(...)


chenl:
So, it fails on windows.
Closing this PR, will try to resolve the windows issue.

PR #1314, opened 2016-04-30
issue#1310

chenl:
undo PR#1194


niphlod:
yep, it needs to be reverted but needs additional love too....

PR #1317, opened 2016-05-03
more tests for validators

leonelcamara:


leonelcamara:
Scheduler tests randomly failing on travis this time for py2.6....

mdipierro:
This does not pass python 2.6 tests. I am ok with dropping support but then we should remove those tests too.

leonelcamara:
@mdipierro it passes 2.6 tests, this was just the scheduler tests randomly failing in Travis as they sometimes do

PR #1318, opened 2016-05-04
Check if host is denied before verifying password

leonelcamara:


mdipierro:
Thank you Leonel and Narendra.

PR #1326, opened 2016-05-09
make allows_jwt a real decorator. Tests included!

michele-comitini:
Makes allows_jwt behave as documented.

@niphlod otherwise argument is necessary?


leonelcamara:
Nice, a fix **and** some tests for jwt!

niphlod:
Kudos for @mcm indeed

PR #1339, opened 2016-05-27
few lib2to3 fixes

ilvalle:
A lot of  lib2to3 fixes.


niphlod:
I'd highly recommend avoiding tinkering with external libs (the one under contrib) which are imported by projects of their own (pypyodbc, pymsql, etcetcetc).
The whole "while 1" replaced with "while True" kinda bugs me. py2 performances with the former are far better than with the latter and py3 should accept while 1 without issues to my knowledge

BuhtigithuB:
Second Simone about changing contrib/ I wouldn't touch those except the one
that are "own" wrote by Massimo or Simone (like Redis Cache or Markmin) as
to my knowledge don't have other repositories...

ilvalle:
I agree with both of you regarding contribs, I had the same idea before starting running the fixes.
However web2py's contrib are the facto part of the project since are really rarely updated, as a result in this first step I decided to update everything but pydal (which needs few fixes as well).
@niphlod what about the failed test regarding scheduler, is it due to this pr?

niphlod:
I don't see any failed test...

Anyway, web2py is not in charge of those modules (they are in "contrib" for a reason). 
Raise a PR on their own repo (or see if they did already taken the 2to3 path) because we need to update their version first and then pull back into our own copy.

BuhtigithuB:
We could contribute back to these lib and propose your update as it is
done... Though it may lead to merge hell on their side depending the amount
of change... But for instance feedparser 31 lines very minors changes, I
think a PR would be accept easily...

I can handle that if you want and update the contrib once the PRs are
accepted and new releases are made?

ilvalle:
So, which contrib should I touch?
Updating only certain contrib and other no can be a mess, web2py has contrib un-updated for years and the updated version (if exist) can be no longer compatible, we can try but it will be a very long path to follow.

see failed tests here travis-ci.org/ilvalle/web2py/builds

BuhtigithuB:
I had updated a couples of the external contrib "recently" : feedparser (is
the last up to date version), markdown2, PyPyODBC,

Need attention (I don't remember having updated them) : rss2, ipaddr, pam,
pbkdf2, qdb (not sure there is a external repo), timecollect (seems we can
deleted as it seems for some compatibility issue with python 2.6),
 gae_retry (seems external not sure), aes, DowCommerce (not sure if
external), AuthorizeNet (seems external but it states it had been
modified), minify (external but modified), pyrtf, pymysql,
 pysimplesoap, pyuca, pg8000, memcache, login/(not all but
many), gateways/fcgi, fpdf

I think I traverse everything...

I would suggest that we classify/identify each contrib as internal or
external at the top of the file in order to not have to investigate if
there is a external repository or not... We can't do that for external as
we can't modify them though (or we need to move them into a subfolder). But
for the internal contrib, there is many that are not properly identify
author, licence, etc. I know there is some library that are really short,
but it would make think easier to maintain IMHO...

ilvalle:
Which contrib should be left untouched by lib2to3 fixes?

niphlod:
better. I'd be extra-careful with those autorewrites happening on 

  gluon/contrib/shell.py
  gluon/restricted.py
  gluon/tests/test_serializers.py
  gluon/tests/test_tools.py
  gluon/validators.py

IMHO they need to be rectified by a human brain.

ilvalle:
ok, I removed the fixes lib2to3.fixes.fix_numliterals and lib2to3.fixes.fix_types, would you review them separately?

niphlod:
it was pretty obvious that it was borked (if not blabla in (type, type)). I'd have toggled them only on the beforementioned files and then fixed manually.
IMHO there's no point in having a huge PR doing lots of stuff on any number of files and then leaving nonetheless the code not py3 compatible. I'd just work on this (and then rebase) to keep one huge commit to "blame" in case something doesn't work.

BuhtigithuB:
Good point Simone...

ilvalle:
can others contribute to this pr?

leonelcamara:
@ilvalle Sure we can contribute to ilvalle:py3_fixes branch and it will then show up in this pull request.

ilvalle:
a couple of issues:
- relative imports don't work well with the current approach to run system tests
- which is the best way to handle correctly on both py2 and py3 ```ur'\x%02x' ```? on py3 is syntactically wrong.

leonelcamara:
As far as I can tell the ur prefix is only used in utf8.py which if I'm not mistaken is then only used by languages.py wouldn't we be better off just removing the utf8 module and adapting languages.py to not use it. I can't really understand why it needs the utf8 module but maybe someone can explain it.

niphlod:
@ilvalle : the ultimate goal of relative imports would make - again, theoretically - possible to avoid messing with sys.path and avoid name collision with other modules/packages (e.g. every once in a while something pops up with "utils.py" which is a pretty common name).
Last time we tried with relative imports there were a few circular imports that needed to be fixed, and then the whole thing was reverted in a rush to release a new version.
Inter-core imports should be relative to make it more self-contained. At some point web2py.py should stop chdir()ing and messing sys.path.
Technically I don't see an issue per se in rewriting tests too to use absolute imports (as far as ci is concerned, there are no "gluon" installed in site-packages), but again, theoretically, gluon can/should be treated as a real package, so using "from gluon.storage import Storage" instead of "from storage import Storage" without fix_sys_path would be a better pick. The only drawback is for people that for whatever reason have "gluon" importable from somewhere else (and in that case, without messing sys.path, the "system-wide" would prevail on the "local/parent" folder).

niphlod:
to everyone: sorry for being pedantic but I'm trying to step into what @mdipierro did (but ATM seems to not have time) of being multiple impersonations of a SteCo that is hugely missing from time to time (a "recorded history" to avoid repeating the same errors, a "code quality reviewer" to avoid backward compat breaks and bring new features in, a "remember what we figured out in the last year" to avoid incorporating nightmares, a "things we should do" to have a somewhat clear path of the future and lastly my usual cranky "I wish I could help more but I really don't have the time"). 
In my days as a major contributor I'd have killed anyone nitpicking my commits (yep, I'm sensitive/touchy) but it was an order of magnitude better than having no code review at all.

applications/admin/private/hosts.deny
```diff
@@ -0,0 +1 @@
+127.0.0.1 2 1464635785
```
niphlod: this should be emptied

mdipierro:
Let's step back and think a moment what we are doing. Do we want to make web2py running under python3? Did we ever agree to do that? We agreed for pydal but not for web2py. Do we really want to do that? Isn't our time better spend making web3py work with python 3 and forget about all those dependencies? Anyway, that said. If we really want to port web2py to python 3 than some of the stuff in contrib needs to be ported. We cannot ship a half baked solution where some stuff like logins and payments not working. Not everything in contrib is well maintained by third parties so if we have to fork it so be it. Some stuff is already forked (pysimplesoap, pyrdf, eas, AuthorizedNet). Some stuff like pg8000 should just be removed in the process.

niphlod:
if web2py is lacking manpower, web3py lacks everything. IMHO if you/we want to push for web3py there's a clear lack of directions (and some bad directions that we're trying to steer away in web2py already (ci, coverage, docs, tests, roadmap, etc)). 
web3py is 21 commits not updated in the last three months. No tests, born to be something to work on and then backport TO web2py, right now it misses fixes contributed to web2py that needs backporting to web3py. It has close-to-none "shiny-new-features", nor it addresses long-standing ones of web2py (maybe it's just a lack of the aforementioned roadmap).

back on 2to3ing. we didn't discuss it. But seems to have at least or 3 peoples interested in it, why not? contribs????? Personally I don't use a LOT of contrib for the same reason (and some additional, already explained in detail) you identified: untested fork or undocumented features. If this is a first step towards aligning core-web2py TO python3, I won't discourage it. Let unmaintained third parties module die as soon as py2 is. They are already dying by lack of usage/maintenance.
IF web3py markets as a something that with a few changes can run web2py apps, it means that web3py will target 2.7 and 3.x, and so will NEED the same 2to3ing process. ATM it isn't a py3 project adapted to work on 2.7 too, it's the other way around, just as web2py is right now.
Marketing web3py as "let's spend time on it" without even a usable beta won't solve the issue that you'll have with contribs: if AuthorizeNet is NEEDED, web3py will NEED to spend time 2to3ing it. If web3py will leverage the manpower people are putting in web2py, why not ?

leonelcamara:
I'm sort of against web3py, I think web2py just needs to drop the promise of backwards compatibility from time to time, we should still strive to maintain backwards compatibility as much as possible of course, and we should be loud and create migration guides when we break it (having to create this documentation should be motivation enough for anyone thinking of breaking backwards compatibility to think twice).  
  
Calling a backwards incompatible web2py web3py, and then focusing our efforts on web3py is the same thing as breaking backwards compatibility, it's almost a move a politician would make to have plausible deniability.

ilvalle:
@mdipierro I see this PR as a way to get web2py closer to py3. 
Afterwards, I'm looking forward to seeing the great features introduced in web3py merged in web2py. On the other hand few modules (i.e., CRUD) should be dropped breaking the backward compatibility.

mdipierro:
I do not oppose to 2-to-3 for web2py core. It may be a good thing. I do not oppose 2-to-3 for contrib although we should try to propagate the changes upstream to the original projects. I just think that this time would be better spent on building a successor of web2py, keeping what we want to keep and making it leaner. The reason I say it is because the effect of porting of web2py to python 3 will break backward compatibility anyway, no existing code will work with it, it will not provide enough added value for existing users to switch (we have seen this with python 3 everywhere). So if we want to attract new users we need some major changes (I know what I want, we do not all want the same thing). So, I do not oppose, and will approve your patches as long as they do not break anything by my interest lies elsewhere.

leonelcamara:
>  and will approve your patches as long as they do not break anything

Well we want to break some of the stuff otherwise this is useless. We also want to make web2py leaner and better. If your plan for web3py is that it will run web2py applications with minor modifications, why not evolve web2py itself?

ilvalle:
I've fixed and enabled few tests for py3.
The main change I've introduced is in html.py now ```xmlescape``` returns bytes instead of str. This is because I haven't found a way to return a string containing ```\xc3\xa0\xc3\xa9\xc3\xa8\xc3\xbb\xc3\xb4\xc3\xa7```
which is the representation of ```'àéèûôç'```. For PY2 this change is transparent.
I great step now would be working on scheduler and language. Fell free if anyone want to contribute, I can't for a while.

BuhtigithuB:
Can we make this patch not backward incompatible in the sens that what
actually work still work? I mean not breaking anything but instead
deprecated old thing in favor of new (no backward warranty) "tools"? We
already discuss of that I mean instead of trying to maintain old
(broken/unmaintainable) stuff, just write new thing beside like Forms in
W3... So we don't mess with SQLFORM we instead write a new form processing
tool which is not "stable" (backward compatible) for now... So someone with
a huge app can make it evolve slowly. Can we have the same code base
execute in python 2.7 and 3.4+ python? Do we need the old code to execute
under both version? If we don't try to port the old code into py3 but state
that it only work properly in py2.7 and make sure this code only can
execute in py2.7 or py3... Sure certain piece of code need to work in both
interpreter... Can we identify it, then work to make compatible and
backward compatible... All mean, i think we need a plan on how we
transition to python 3. Massimo's want to drop w2p in favor a something
different without backward warranty... But I think this (even if I dream to
see w3py soon) is not going to be a good answer to web2py all member of the
w2p user base. Which will get many chances of splitting the user base into
w2p and w3p... So the question is do we want to make evolve w2p or not? If
we drop it what will be the path of transiting or actual apps into w3p? IRC
talk about that?

mdipierro:
Before I merge this. If one is still using python 2.7, what will break?

niphlod:
Halt! 
IMHO a few are missing pieces. 
2to3 here ( @ilvalle correct me if I'm wrong ) just nears the gap between 2.7 and 3, it doesn't break anything (given auto-tools are always something to be watched over, as noted in https://github.com/web2py/web2py/pull/1339#issuecomment-222459255 )
Given 2.6 dropped, writing (or re-organizing) code that works in 2.7 and 3 is easier, and this is one of the steps to accomodate a web2py that WOULD work with a 3.x interpreter (obviously the app too needs to be coded in 3.x syntax). And in fact, 2.7 tests are not breaking.

niphlod:
in addition to the above statement: given coverage is low (0% in contrib) everything not covered MAY break (i.e. it's not sufficient to rely on travis-ci and appveyor telling "tests are passed"). 
That's why I was advocating towards this first step NOT touching contrib/* if there are no users testing those contribs with the "new syntax".

niphlod:
also: this has become a gigantic PR that I'd fear to merge as it is without 1-by-1 approach.

- gluon/contrib/shell.py IMHO is wrong
- I don't get the pycrypto dependency

That being said:
- a lot of contribs can surely be stated as non-breakable (e.g. authorizenet.py changed only prints, so I'm 99% sure that it won't break) even without tests
- others needs careful inspection (if someone didn't already, e.g. those pesky "long" removals in ipaddr.py and fcgi.py)
- some are nonsensical, i.e. x509_auth.py, which relies on m2cypto, which has NO py3 package... it surely won't break for 2.7 but it has no sense either way
- skipping tests for py3.x does more harm than good in "let's be sure something will work on 3 the same way it worked on 2" but they won't affect 2.7 functionalities, so for the moment those are okay (maybe a "FIXME" to remember they need to be addressed in the future)
- rewriting tests to make them pass forcefully on 3.x IMHO should be avoided at all costs (see removal of self.assertEqual(json(1), json(1L)) from test_serializers.py. Small thing, but hey, that is cheating! )

@ everyone contributing to this: a hell of a good work, (me being nitpicking and all) I only found the aforementioned issues :-P

ilvalle:
@niphlod I've already added a few ```# FIXME PY3``` here and there I will try to add more, ```self.assertEqual(json(1), json(1L))``` in py3 is ```self.assertEqual(json(1), json(1))``` as a result I removed it, I will re-add it for py2.
pycrypto dependency is mandatory since the current contrib is not py3 ready, nowadays the *battery included* feature could be discussed a little be more.
@mdipierro this pr passes all py2 tests but they don't cover everything.

leonelcamara:
@ilvalle can you give me write access to your web2py fork so I can contribute to the py3_fixes branch, I'm kind of a git newb and this would make it a lot easier.  
  
What I want to add right now are some fixes to make cache disk windows work on py3 too and an updated AES contrib since upstream pyaes has a py3 compatible version now

niphlod:
imho we should hold a step and merge some things to get down the commit revision: merging is going to be hell on those sizable commits and number of files...nobody needs to have everything working on 3.5 just after this commit...

ilvalle:
@leonelcamara I gave you access. However I agree that we should split this porting in several steps.
In addition, there are few decisions to take mainly related to unicode string. 
For example, in py2.7 ```IS_LENGTH(2)(u'°2')``` returns ```('\xc2\xb02', None)```
In py3 isn't better to return the unicode string received in input without encoding it?

leonelcamara:
@niphlod @ilvalle 

Ok I will hold on making my changes and simply redo them once this pull request is done.

I think that validator is working correctly, what matters should be the length in bytes if ultimately you want to store it in a field of limited size in the database. However it should probably have an option to just count the number of characters in the unicode string instead of the number of bytes it takes once encoded.

niphlod:
I did not expect this development (meant not as "code development" but as "history development") on the PR but I think it's safe to assume that it's better to STOP adding commits, 
wait for reviews (hoping to be not the only one) and then wait a few days for @mdipierro to merge it (let's say on the 13th ?).

Seen that nobody told me I was wrong on the big picture side of things, this can be considered as the first of a few steps to run web2py on py3 (for the time being, and for the remainder of this reply, let's consider only 3.5)

IMHO there'll be a bit more to wait for the next release, because basically everything changed. We missed the agile dogma "release small and often", 
and possibly the commit will receive a lot of blames but we don't need to make a huge deal about it. 
It's just how it turned out to be. Discarding would mean loosing a good shot at making web2py py3 friendly.
In total, I think I spent roughly four hours checking diffs in the past days and most of the things I noticed have been fixed. I just spent another two reviewing the current state in a single breath, 
so hopefully this would be the last ^_^

@ilvalle (and/or other contributors): we need a detailed report of the things that needs further consideration (possibly with a thread on web2py-dev) to let things
be sorted out in the short-medium term. I'd urge to be "better safe then sorry" so feel free to report things that are working but you're not satisfied with (we all know the 
entire codebase has some obscure code paths). I'd vouch even for a "gotchas-style" bullet list, that we can use as a baseline for a milestone where we can assume people can start testing py3 (not only the brave ones)
Also, future merges would need a more careful "inspection" to avoid re-inserting py3-not-compatible syntaxes. I'm not talking about a code-review or a board or a guide on how to write py3 safe code, just... let's be careful and not
waste this effort also in the future.

My remaining concerns (you all can safely tag them all as "IMHO", so not set in stone):
- gluon/cache.py ( for key in list(storage.keys())) # why consolidating it in a list ? a more readable "for key in storage" would be a better pick, and will possible enable leverage future optimization in python
- contrib/fpdf* hopefully it's exactly the 1.7.2 as latest stable in its own repo
- gluon/contrib/gateways/fcgi.py those changes IMHO will break something in 2.7 . Given the rise of uwsgi and scgi we should instead rely on the new (and supported) fcgi which in turn depends on flup. A fat warning stating that users relying on fcgi would need flup (which for python/fastcgi hosters is ALWAYS included) would seal the deal
- gluon/contrib/login_methods/ldap_auth.py "if type(allowed_groups) != type(list()):" got translated to "if not isinstance(allowed_groups, type(list()))" . Should instead be "if not isinstance(allowed_groups, list)"
- gluon/contrib/login_methods/x509_auth.py won't work with py3, another method is needed
- gluon/tools.py : again, m2crypto is not available for py3
- gluon/contrib/shell.py smells funny, but this would be the 312098312 time we say web shell should not be used. Next release would probably be a REEEEEEALLY good time to drop from the admin app and let the contrib module with a big "won't ever FIX"
- gluon/contrib/websocket_messaging.py won't certainly break 2.7, but I'm not sure it can be considered py3 safe
- gluon/globals.py given the importance of the module, remember to put those #FIXME for py3 in the "let's discuss" recap (or clarify what is left to be decided)
- gluon/highlight.py nothing breaks 2.7 but cgi.escape should be addressed for 3.x
- gluon/languages.py I'd drop iterkeys() alltogether: it doesn't have any improvement over keys() or just the "dry" "for key in dict" syntax
- gluon/recfile.py it has to do with the initial design concept, nothing will break 2.7 but we should really be concerned about relying on auto-closing the stream (py3 will emits warnings about it)
- gluon/shell.py as contrib/shell.py: to me those replacement are not 1:1 . That being said it's on the doctest part which is by far the least utilized part of shell.py

- here and there: "from gluon" instead of "from .gluon": cheating circular imports or forgot to rewrite ?	

I'm curious about adding  current.response.static_version = None on test_router : did some behaviour change ?

on the "discuss not here" I'd also put:
- DAL routine tests should be fixed
- IS_IP* validators could use the ipaddress library (mostly because it's included in 3 and there is an official backport for 2.7) instead of ipaddr-py from google

On the CHANGELOG should also be reported as the very first warning that web2py DOES NOT SUPPORT ANYTHING BUT 2.7 (i.e. not only we don't officially support 2.6 but also
we introduced changes that will surely break running on 2.6)
We may consider adding a CHANGELOG entry for brave users to test 3.5 and report back findings.

PS: @ilvalle I didn't get why pycrypto is needed. What startles me is that AFAIK it's accessible as "Crypto" but I don't find it in any piece of "import Crypto" on the diff. Feel free to bash me publicly or in private :P

gluon/_compat.py
```diff
@@ -0,0 +1,106 @@
+import sys
```
BuhtigithuB: Why are we duplicate this file? 

https://github.com/web2py/pydal/blob/b5f91ba4837da33f1f3036fa5a821a28dee68693/pydal/_compat.py

gluon/scheduler.py
```diff
@@ -41,7 +41,7 @@
 from gluon.scheduler import Scheduler
 
 def demo1(*args,**vars):
-    print 'you passed args=%s and vars=%s' % (args, vars)
+    print('you passed args=%s and vars=%s') % (args, vars)
```
BuhtigithuB: I find this cleaner : 
print('you passed args=%s and vars=%s' % (args, vars))

Or even : 
print('you passed args={} and vars={}'.format(args, vars))

gluon/_compat.py
niphlod: because pydal is a separate project and it may change it.

2016-06-07 16:00 GMT+02:00 BuhtigithuB <notifications@github.com>:

gluon/tests/test_scheduler.py
```diff
@@ -309,12 +309,12 @@ def testRepeats_and_Expired_and_Prio(self):
         self.db.commit()
         self.writefunction(r"""
 def demo1(*args,**vars):
-    print 'you passed args=%s and vars=%s' % (args, vars)
+    print('you passed args=%s and vars=%s') % (args, vars)
```
BuhtigithuB: Same thing here : 

I find this cleaner : 
print('you passed args=%s and vars=%s' % (args, vars))

Or even : 
print('you passed args={} and vars={}'.format(args, vars))

gluon/contrib/redis_scheduler.py
```diff
@@ -31,7 +31,7 @@
 from gluon.contrib.redis_scheduler import RScheduler
 
 def demo1(*args,**vars):
-    print 'you passed args=%s and vars=%s' % (args, vars)
+    print('you passed args=%s and vars=%s') % (args, vars)
```
BuhtigithuB: I find this cleaner : 
print('you passed args=%s and vars=%s' % (args, vars))

Or even : 
print('you passed args={} and vars={}'.format(args, vars))

gluon/contrib/markmin/markmin2pdf.py
```diff
@@ -20,7 +20,8 @@ def removeall(path):
     def rmgeneric(path, __func__):
         try:
             __func__(path)
-        except OSError, (errno, strerror):
+        except OSError as xxx_todo_changeme:
```
BuhtigithuB: Can we properly fix this now? If not we need a proper Fix me/todo comment...

gluon/custom_import.py
```diff
@@ -86,7 +86,7 @@ def custom_importer(name, globals=None, locals=None, fromlist=None, level=-1):
                         try:
                             result = result or sys.modules[modules_prefix+'.'+itemname]
                         except KeyError as e:
-                            raise ImportError, 'Cannot import module %s' % str(e)
+                            raise ImportError('Cannot import module %s' % str(e))
```
BuhtigithuB: Did we address byte/string/unicode of py2-py3? If so, is str(e) here will cause issue?

gluon/tests/test_web.py
```diff
@@ -11,12 +11,12 @@
...
 
-from fix_path import fix_sys_path
+from .fix_path import fix_sys_path
 
 fix_sys_path(__file__)
 
-from contrib.webclient import WebClient
-from urllib2 import HTTPError
+from gluon.contrib.webclient import WebClient
+from gluon._compat import urllib2
```
BuhtigithuB: Why keeping urllib2 in _compat and not use urllib? The objective is not to drop it once it is not required anymore (if ever possible once it need we drop support for 2.7 which may take time...)??

BuhtigithuB:
This is massive change!! 

I pretty much give a look to every changes an leave questions or comments...

I would say that these changes can be applied alone as they are not strictly speaking py2/3 transition related : 
running lib2to3.fixes.fix_idioms

But I guess they would cause merge hell if apply before, but if they can be apply after without issue they are good candidate to be postponed if we want to reduce the size of this PR and keep only expressly p2/3 transition stuff inside this PR...

ilvalle:
@BuhtigithuB I've fixed the print, the other suggestions can be applied after the merge.
@niphlod thanks for the review.
- ok for ipaddress, see also my comment on test_validators.py:1031
- ```shell.py``` I totally agree, as a matter of fact cloud environments (as pythonanywhere) have better features. In addition I propose to remove it from py2.7 too.
- ```languages.py``` it depends on utf8.py. here a web2py guru is needed to port it to py3, my hope for py3 is to use native functions without that dependency.
- uwsgi/fgci, nowadays is it visible to support only the former?
- cache.py ```for key in storage``` is not valid in py3, it raises "dictionary changed size during iteration" due to the ```del``` statement in the loop
- in this first step I've introduced pycrypto because the current contrib/aes.py isn't py3 compatible; but it can be updated with https://github.com/ricmoo/pyaes

@leonelcamara while you are working with cache, since travisci supports both memcache & redis, it would be nice to have tests for them too; it will make easier and safer to update their libraries to py3. Redis can be used for unittesting scheduler_redis too. @niphlod what do you think?

Ok for waiting few days in order to double check that py2.7 works correctly.

niphlod:
perfect ilvalle, now I understand the pycrypto req. Although for now it should be on the "lets consider it in another PR" column. 
Also, leave scheduler and scheduler_redis out of further modifications, as I need to push a new version with a new feature. Tests on them too, as they are spinned up in an intricate way, but alas, I couldn't find any better way.
Given I'm the only supporter of a Trello board, I'll recap here points to be discussed in an hour or so. Frankly my head hurts ^_^

niphlod:
voilà. https://github.com/web2py/web2py/issues/1353 and https://groups.google.com/forum/#!topic/web2py-developers/wyG4x_-Gxb4 created. Feel free to bash yours truly for keeping things in order.

gluon/utf8.py
```diff
@@ -10,12 +10,15 @@
...
+from __future__ import print_function
+from gluon._compat import builtin as __builtin__, unicodeT, iteritems
+
 __all__ = ['Utf8']
 
 repr_escape_tab = {}
-for i in range(1, 32):
-    repr_escape_tab[i] = ur'\x%02x' % i
+#FIXME PY3
+#for i in range(1, 32):
```
mdipierro: why is this commented?

mdipierro:
First of all congratulations for this PR which was clearly a lot of work. I am ready to merge it, as soon as I get an answer to my comment. Why was this line in utf8.py commented?
 -    repr_escape_tab[i] = ur'\x%02x' % i	
Is it no longer necessary? I do not see it breaking anything in py3.

gluon/utf8.py
ilvalle: It raises a syntax error on py3 and I commented it out while working. It should be re-enabled before releasing a new stable for py2.7.

leonelcamara:
@mdipierro python3 doesn't have ur'' strings.

mdipierro:
yes but we can use: repr_escape_tab[i] = unicode('\\'+'x%02x' % i) I think.

ilvalle:
@mdipierro you can merge this PR. @niphlod do you agree?

mdipierro:
And it is done!

BuhtigithuB:
Great work!!!

mdipierro:
Yes. Congratulations everybody. This was a lot of work!

PR #1341, opened 2016-05-28
Ignore internal attributes when checking whether new session was changed

nextghost:
The simple hash check always fails for new secure sessions because the `_secure` flag is added *after* calculating the initial `response.session_hash`. Therefore new secure sessions will be stored even when they're effectively empty. This patch makes sure that a new session contains something *other* than purely internal attributes before it is stored.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -1136,6 +1136,12 @@ class Session(Storage):
         return True
 
     def _unchanged(self, response):
+        if response.session_new:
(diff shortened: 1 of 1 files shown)


nextghost:
Also, why does `_try_store_in_db()` force storing of new sessions even when they're empty?

mdipierro:
@nextghost. If it does it is a bug. Will check but feel free to submit another patch for that.

PR #1359, opened 2016-06-13
Don't truncate texts on SQLFORM.grid HTML Export

### LINKED ISSUE #1358: Don't truncate texts on SQLFORM.grid HTML Export
Currently ExporterHTML truncates text fields longer than 16 characters, even if maxtextlength attribute is specified in SQLFORM.

Make ExporterHTML render the whole texts, just as other format exporters do (csv, json, etc).

oscarfonts:
Allow HTML export to display long texts.

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -2476,7 +2476,8 @@ class SQLFORM(FORM):
                 response.headers['Content-Type'] = oExp.content_type
                 response.headers['Content-Disposition'] = \
                     'attachment;filename=' + filename + ';'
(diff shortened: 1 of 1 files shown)


niphlod:
this still leaves the same issue when individual column length is passed... IMHO a better fix would be to not truncate the HTML export at all, in any case.

oscarfonts:
@niphlod agreed. Not truncating texts at all means cleaner code, and it's what users would expect in an export, I guess. So, I'm reconsidering the PR.

niphlod:
perfect

PR #1387, opened 2016-07-05
enabled py35 unittests on win

ilvalle:
fixed minor issue with cacheOnDisk

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,11 @@ environment:
   matrix:
     - PYTHON: "C:/Python27"
       COVERAGE_PROCESS_START: gluon/tests/coverage.ini
+      PYTHON_ARCH: "64"
+
+    - PYTHON: "C:/Python35"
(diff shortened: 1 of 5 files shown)


niphlod:
perfect. I wouldn't know about the utility to test 32 AND 64 archs in 2016 ... in any case paths to python should be either all with "\" or with "/".
Also, what about using https://github.com/Legrandin/pycryptodome instead of pycrypto ???

ilvalle:
Which arch would you test?

niphlod:
to avoid augmenting the matrix (and since I think Travis-CI has only 64 bit, too), I'd go for 64 bit only. @mdipierro, thoughts ? It may also be the time to package as a binary with pyinstaller rather than bbfreeze only 64 bit and/or think for the next release advertising py3 support a py3 binary.

mdipierro:
@niphlod, I agree. @ilvalle, can I merge or should I wait for changes based on Simone's recommendations?

ilvalle:
Updated as suggested, @mdipierro you can merge it.

PR #1397, opened 2016-07-18
Massive memory leak in Scheduler with PostgreSQL

nextghost:
When you run Web2Py scheduler on PostgreSQL, you'll get massive memory leak to the tune of 1GB RAM consumed in 48 hours. The problem is not really a bug as such but rather a horrible design mismatch between PyDAL and the pg8000 driver.

Long story short, pg8000 driver supports multiple user-friendly placeholder formats for query arguments and converts them to the format supported by PosgreSQL server. It also expects that the number of unique query strings will be fairly limited (because the arguments will be passed separately) and stores the converted query strings and some other computationally expensive metadata in cache.

Meanwhile, PyDAL happily dumps all arguments into the query string itself, which means that every single scheduler heartbeat will turn into a unique query string and bloat the pg8000 cache.

This patch is a dirty workaround that disables pg8000 query cache. The correct solution would be rewriting PyDAL to pass query arguments through placeholders. Otherwise you're giving up a pretty big performance boost.

### DIFF
--- a/gluon/contrib/pg8000/core.py
+++ b/gluon/contrib/pg8000/core.py
@@ -1568,7 +1568,9 @@ class Connection(object):
             statement, make_args = cache['statement'][operation]
         except KeyError:
             statement, make_args = convert_paramstyle(paramstyle, operation)
-            cache['statement'][operation] = statement, make_args
+            # FIXME: Rewrite PyDAL to generate query strings with placeholders,
+            # then uncomment the line below
+            #cache['statement'][operation] = statement, make_args
 
         args = make_args(vals)
         params = self.make_params(args)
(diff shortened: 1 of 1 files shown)


niphlod:
once again, pg8000 falls short. this should be reported upstream and fixed/tuned/overridden properly instead of us maintaining a patched version

nextghost:
@niphlod This is a temporary workaround for PyDAL design problem. There is noting here worth reporting to upstream pg8000 maintainers. Code that triggers this problem is by definition generating awfully bad SQL queries.

mdipierro:
I agree with Niphlod. This should be considered a bug in pg8000. I should be able to send any queries I want with or without placeholder and I would not have a memory leak.

nextghost:
Well, never mind. I've just found out that even with query cache disabled in the pg8000 driver, the PostgreSQL server itself still gets flooded with prepared statements and eventually runs out of memory. And I've also noticed that the pg8000 driver leaks prepared statements whenever your run CREATE TABLE or ALTER TABLE so now there's a reason to report this upstream...

PR #1399, opened 2016-07-18
fixed most of py3 warnings, output is much cleaner this way

niphlod:
the same thing should be done with pydal...


niphlod:
whoops. resending after typo (and rebase)

PR #1400, opened 2016-07-18
fixed most of py3 warnings, output is much cleaner this way

niphlod:
same should be done with pydal (and (or?) a better "import" of pydal unittests)


niphlod:
umpf. need to fix appveyor, even if this got tested fine on my rig which should be 1:1 with appveyor env for 3.5 ... keep on hold.

PR #1402, opened 2016-07-19
this crashes win and 3.5

niphlod:


niphlod:
our portalocker implementation crashes on py3.x . I'll try to port over either 

https://github.com/WoLpH/portalocker

or

https://sourceforge.net/p/roundup/code/ci/default/tree/roundup/backends/portalocker.py

PR #1404, opened 2016-07-20
add 'form-control' to _class of autocomplete input

remcoboerma:
This way it looks better because of bootstrap3 integration. 
http://getbootstrap.com/css/#forms

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -736,7 +736,7 @@ class AutocompleteWidget(object):
         if self.is_reference:
             key2 = self.keyword + '_aux'
             key3 = self.keyword + '_auto'
-            attr['_class'] = 'string'
+            attr['_class'] = 'string form-control'
             name = attr['_name']
             if 'requires' in attr:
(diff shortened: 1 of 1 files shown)


mdipierro:
Thank you. I would like a second opinions about this PR. I understand the purpose but I really do not want to tie web2py more to bootstrap. If anything I want to untie it.

mdipierro:
So no because you can do:

```
SQLFORM.widgets.autocomplete._class = "string form-control"
```

and add any class you want.

remcoboerma:
Thanks for that. But does untying break backward compatibility? It all depends on what one would consider backward compatible i guess. Don't get me wrong, it would be great to get web2py frontend agnostic but still easy enough to work with all different kinds of these frameworks.

mdipierro:
Working (slowly) for some time for a new library to replace SQLFORM. We are not removing SQLFORM but it has unfixable problems. The new one will be similar but not backeard compatible and have an easier way to couple css to the form.

PR #1405, opened 2016-07-23
Master

sollua:


mdipierro:
I am confused by this PR. It includes two css files in admin but does not appear to use them...

niphlod:
me too, @sollua please explain what you wanted to do. unless a reply pops up, we're going to discard this PR

mdipierro:
I think this applies to admin2, not master.

PR #1410, opened 2016-07-28
fixes #1403

### LINKED ISSUE #1403: portalocker crashes py3
need to integrate 
https://github.com/WoLpH/portalocker
or
https://sourceforge.net/p/roundup/code/ci/default/tree/roundup/backends/portalocker.py

with our own (that has "specialized" GAE "additions")

niphlod:
and yes, we don't have pywin32 dependency anymore!!!!

@gi0baro : you can fetch portalocker.py and tests as they are and include in pydal, no changes needed, except for the line importing the _compat module. If you agree, I can push a PR to pydal's repo containing these changes

### DIFF
--- a/gluon/portalocker.py
+++ b/gluon/portalocker.py
@@ -26,16 +26,35 @@ Methods::
 
 Constants::
 
-   LOCK_EX
-   LOCK_SH
-   LOCK_NB
+   LOCK_EX - exclusive lock
+   LOCK_SH - shared lock
+   LOCK_NB - don't lock when locking
+
+Original
+---------
+http://code.activestate.com/recipes/65203-portalocker-cross-platform-posixnt-api-for-flock-s/
 
 I learned the win32 technique for locking files from sample code
 provided by John Nielsen <nielsenjf@my-deja.com> in the documentation
 that accompanies the win32 modules.
 
 Author: Jonathan Feinberg <jdf@pobox.com>
-Version: $Id: portalocker.py,v 1.3 2001/05/29 18:47:55 Administrator Exp $
+
+
+
+Roundup Changes
+---------------
+2012-11-28 (anatoly techtonik)
+   - Ported to ctypes
+   - Dropped support for Win95, Win98 and WinME
+   - Added return result
+
+Web2py Changes
+--------------
+2016-07-28 (niphlod)
+   - integrated original recipe, web2py's GAE warnings and roundup in a unique
+     solution
+
 """
 
 import logging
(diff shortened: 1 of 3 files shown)


niphlod:
BTW: wait before merging this. portalocker is tested, but not entirely in every way it is used in web2py. it may need a bit of additional love.

niphlod:
ok, triple funny moments here ....

recap: we have two portalockers in web2py. one in gluon.contrib and one in pydal.contrib.
First funny thing: py3 tests on win crash because on test_cache there's a test which defines a table... for some reasons two portalockers can't coexist, or at least it seems that the latter overrides the former...
Second funny thing: replaced portalocker in pydal with the same from this PR, same error applies, PLUS in a shell as soon as a table is defined.
Third funny thing: if I change names of the modules (i.e. to avoid weird overrides), nothing changes... as soon as the one from pydal is called, exceptions are thrown
Fourth funny thing: everything works exactly as expected (i.e. no issues whatsoever) if pydal uses portalocker from gluon.contrib instead of its own.

@gi0baro , @mdipierro , @michele-comitini , @ilvalle , @abastardi : thoughts ?

mdipierro:
scratching head... do you have pip install pydal?

niphlod:
nope, I use virtualenvs and triple-checked.... 
the big baseline is that apparently this portalocker works fine ONLY if it's the only one being imported. and only in 3.5 on win. test_cache definitely blocks because there's a weird thing going on on __wait_portalock()__ in cacheondisk. AFAIK 

portalocker.lock(val_file, portalocker.LOCK_EX)

blocks indefinitely until a lock can be gained (there are no timeouts)... 
doing that try: except: in that case would serve nothing (but in the specific scenario, swallows the exception of the call on pydal's portalocker and sits there in the loop)

Funny thing unrelated: this "recompilation" of portalocker is quite the exact same as the one django uses.......

mdipierro:
I'd be happy to only have one portalocker, the one in pydal. I understand this does not solve the problem.

niphlod:
well, the way I see it, we have some solutions:
- someone more skilled in windll than me takes a shot at the error, as it probably highlights a deficiency when trying to call the same underlying method from different modules
- try again with pywintypes, which will remain a hard-dependency for web2py on win
- try to use just one module (we'd need to keep pydal's and web2py's aligned) OR given that web2py won't ever work without pydal, use the one from pydal

thoughts ?

mdipierro:
we can use the one in pydal only

niphlod:
OR, but at this point I need multiple eyes to confirm that the included test-suite is fairly complete, ditch this method in favour of just relying on msvcrt.locking alone on win (zope apparently does it)

mdipierro:
Less dependencies the better. I'd do what zope does.

mdipierro:
well. I removed gluon/portalocker (and _compat since I was at it and is already in pydal).

niphlod:
I was just posting that I tried with msvcrt and there was a big miss... no shared lock for reads.

mdipierro:
You should write a blog post about all these issues with locking on windows.

niphlod:
PS: watch out for a lot of failing tests on the current master.

niphlod:
PS2 : once this "move" is complete and doesn't fail, I'll issue a PR with this changes against pydal's contrib.portalocker, then a fix release should be done on pydal, that release tracked in web2py's submodule, and then and only then #1403 will be solved

mdipierro:
Dealt with on the pydal side. Thanks Simone.

Issue #649, opened 2015-01-26
Customize view of disabled apps


niphlod:
_From [paolo.va...@gmail.com](https://code.google.com/u/107767337307670204068/) on March 09, 2013 13:00:56_

What steps will reproduce the problem? 1. Disable an app from the appadmin 
2. Visit the app website
3. you will get the following code:

\<html><body><h1>Temporarily down for maintenance</h1></body></html>

which comes from gluon/main.py:492

raise HTTP(503, "<html><body><h1>Temporarily down for maintenance</h1></body></html>") Please provide any additional information below. The best would be to load a specific view i.e, from applications/app/view/disabled.html

This will give the possibility to customize the disabled view for each application.

_Original issue: http://code.google.com/p/web2py/issues/detail?id=1371_

niphlod:
_From [niph...@gmail.com](https://code.google.com/u/110791369705180298143/) on March 09, 2013 04:43:38_

I'm assuming disabled.html is a static page (i.e. doesn't need templating, extending from layout.html, etc ?)
If so, I can patch it.

niphlod:
_From [paolo.va...@gmail.com](https://code.google.com/u/107767337307670204068/) on March 09, 2013 04:54:07_

Umh, I put the same question to my self too.
My thought is to have something more than a static page,
In a static page you have to define a lot of html code which is already
defined in the layout, for this reason I would propose to have something
more than a static page in which at least I can put {{extent 'layout.html'}}


 Paolo

niphlod:
_From [ABasta...@gmail.com](https://code.google.com/u/103720714265511973715/) on March 09, 2013 05:18:04_

Some other thoughts: https://groups.google.com/d/msg/web2py-developers/guV2od3TV6c/iiyZ9wPq9jkJ

niphlod:
_From [niph...@gmail.com](https://code.google.com/u/110791369705180298143/) on March 09, 2013 06:43:04_

@paolo: the whole point of having the site down for maintenance is not executing any code. If you want to have your page to respond differently (e.g. a desktop and a mobile version), or you're basing your maintenance page on some code execution, than this is a chicken-and-egg problem. You could as well have a maintenance "app" and redirect early in your production app to that maintenance.
On the other hand, I don't see a real limitation on having a maintenance page that is static. Extending a template for the maintenance page, and saving a static version of the result is not hard (ctrl+s in almost every browser). 

[...]

PR #1420, opened 2016-08-08
Custom 503 pages

vinyldarkscratch:
Added support for custom down for maintenance pages on disabled apps to fix issue #649 as suggested by Richard:
https://groups.google.com/d/msg/web2py-developers/guV2od3TV6c/T-7ZJX6dWe
EJ


gluon/main.py
```diff
@@ -390,7 +390,11 @@ def wsgibase(environ, responder):
                                    % 'invalid request',
                                    web2py_error='invalid application')
                 elif not request.is_local and exists(disabled):
-                    raise HTTP(503, "<html><body><h1>Temporarily down for maintenance</h1></body></html>")
+                    if os.path.exists(os.path.join(request.folder,'static','503.html')):
+                        import urllib
+                        raise HTTP(503, urllib.urlopen(os.path.join(request.folder,'static','503.html')).read())
```
ilvalle: This doesn't work on python 3.5, please consider importing urlopen from pydal._compat

niphlod:
please close this PR and issue one just for the issue at hand: updating each language file with the default translations doesn't belong here (and, frankly, isn't needed at all)

gluon/main.py
niphlod: and why using urllib at all when you can read the file  ?!?!?

vinyldarkscratch:
Oh that's weird, it shouldn't have requested to pull my second commit...  Oops, sorry about that!  (Still getting used to Git repositories and how they all work.)

For the use of URLLIB, I have no idea -- I literally just copied the code from the link I mentioned above and made a few slight changes.

gluon/main.py
vinyldarkscratch: (I never saw these until just now...)

@niphlod You actually have a really good point...that's what I get for 4:30 AM code-writing.  XD

PR #1425, opened 2016-08-16
F redirect in admin/default/design (install of plugin)

zvolsky:
Otherwise causes error if plugin install fails, because redirects to default/design without args/appname.


mdipierro:
Thanks!

PR #1427, opened 2016-08-17
pad/secure_dumps/secure_loads update

ShySec:
added [PKCS #7 v1.5 padding](https://www.ietf.org/rfc/rfc2315.txt)
updated secure_dumps to use SHA256
fixed secure_dumps using truncated encryption_key hash
fixed pad ignoring blocksize argument

Supports older secure_dumps files through __secure_loads_deprecated interface (automatically invoked through secure_dumps for data with old data format).

secure_dumps previously truncated the 20-byte SHA1(encryption_key) output to 16-bytes (hexadecimal). Updated secure_dumps now uses entire 32-byte SHA256(encryption_key) output.

### DIFF
--- a/gluon/tests/test_utils.py
+++ b/gluon/tests/test_utils.py
@@ -89,11 +89,11 @@ class TestUtils(unittest.TestCase):
 
     def test_pad(self):
         test_cases = [
-            (16, 'mydata'), # verify data padding and unpad identity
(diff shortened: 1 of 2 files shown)


ShySec:
Finally got web2py to start up with python3 and to verify system tests

PR #1430, opened 2016-08-17
fixed "/app/static" throwing AttributeError instead of HTTP(404)

ShySec:
HTTP requests for `/app/static` currently throw an AttributeError due to attempting a string replacement when the filename is None. Patch raises HTTP(404) before attempting the string replacement if filename is None.


niphlod:
LGTM

PR #1450, opened 2016-09-16
caching read_pyc

ilvalle:
Running ```ab -n 10000 -c 1 http://localhost:8000/welcome/default/index``` on an i5 with SSD
I got (current master)
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    15   23  13.1     18     166
Waiting:       15   23  13.1     18     166
Total:         15   23  13.1     18     166
```
with this PR
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    15   19  10.4     16     148
Waiting:       15   19  10.4     15     147
Total:         15   19  10.4     16     148
```
The gain is ~15% running welcome default controller. @abastardi would you mind computing TechEmpower Benchmarks?

On the other hand since application files can change, such optimization requires a way to erase the cache (maintained in a dict in cfs.py). An idea is to erase such cache in ```compile_application()```. What do you think?

### DIFF
--- a/gluon/restricted.py
+++ b/gluon/restricted.py
@@ -199,7 +199,7 @@ class RestrictedError(Exception):
 
 
 def compile2(code, layer):
-    return compile(code.rstrip(), layer, 'exec')
+    return compile(code, layer, 'exec')
 
 
 def restricted(code, environment=None, layer='Unknown'):
--- a/gluon/compileapp.py
+++ b/gluon/compileapp.py
@@ -581,7 +581,7 @@ def run_models_in(environment):
             if not regex.search(fname) and c != 'appadmin':
                 continue
             elif compiled:
-                code = read_pyc(model)
+                code = getcfs(model, model, lambda: read_pyc(model))
             elif is_gae:
                 code = getcfs(model, model,
                               lambda: compile2(read_file(model), model))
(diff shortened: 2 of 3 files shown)


niphlod:
does the above only target compiled applications ? if yes, the code can't change. On recompiling, erasing the cache seems the ONLY way to make this work

ilvalle:
Right, it targets only compiled applications because I don't see an easy way to erase the cache when source file changes. On the other hand, compiled files should change only through ```compile_application()```, as a result erasing cache in such method would be enough.

abastardi:
This looks good, but why not also cache the `read_pyc` call in `run_view_in`?

Maybe we should use `getcfs` in place of the two calls to `read_file` in `run_controller_in` as well (we already use `getcfs` rather than `read_file` to load non-compiled model files in `run_models_in`).

Regarding clearing the cache, doesn't `getcfs` already check the modified time of the file and re-read it if it has changed?

ilvalle:
Updated PR to use ```getcfs``` in ```run_view_in```. 
@abastardi you are right ```getcfs``` already checks for ```st_mtime``` so that there is no need to clear the cache.

gluon/restricted.py
```diff
@@ -199,7 +199,7 @@ def __str__(self):
 
 
 def compile2(code, layer):
-    return compile(code.rstrip(), layer, 'exec')
+    return compile(code, layer, 'exec')
```
mdipierro: I will take your patch but the rstrip() is important because on windows the web editor has the bad habit of adding the \n\r at the end and compile(..) complains.

ilvalle: Since web2py supports py27+ according to https://docs.python.org/2/library/functions.html?highlight=compile#compile  rstrip is no longer necessary on both mac and win

niphlod: +1 for py27+ once again :D

mdipierro:
As long as you tested ok. I do not have windows to test it.

PR #1461, opened 2016-09-21
fixes #1310 and improves upon ideas shared on #1450

### LINKED ISSUE #1310: compiled app's views aren't executed

niphlod:


niphlod:
@ilvalle : since you have been the man in charge of #1450 can you take a look at this ?

ilvalle:
Can't try right now but it looks ok. Btw don't know what is _TEST function.

ilvalle:
To have the same behavior there is in gae, I'd add a call to getcfs at line 722.

niphlod:
hem, do we WANT the same behaviour as in GAE ?

mdipierro:
The test function allows to run doctests from admin. Wan't this break backward compatibility for compiled apps? I am on with it but we should make a very prominent note of it.

niphlod:
if it's not final, wouldn't have been better to continue the discussion before merging ?

mdipierro:
I will revert if needed. We can still discuss.

niphlod:
NP: I'll open a new issue to keep the track or this will get lost, given the "merged" status is normally hidden

ilvalle:
I meant something like
```
diff --git a/gluon/compileapp.py b/gluon/compileapp.py
index 42794ba..cbd584a 100644
--- a/gluon/compileapp.py
+++ b/gluon/compileapp.py
@@ -724,9 +724,10 @@ def run_view_in(environment):
                                             pjoin(folder, 'views'),
                                             context=environment), layer))
         else:
-            ccode = parse_template(view,
-                                   pjoin(folder, 'views'),
-                                   context=environment)
+            ccode = getcfs(view, view,
+                           lambda: compile2(parse_template(view,
+                                            pjoin(folder, 'views'),
+                                            context=environment), view))
         restricted(ccode, environment, layer)
```

niphlod:
@ilvalle : I went for including cfs directly in template.py . should result in the same thing without overcharging with lambda statements ^_^ . or not ?

ilvalle:
I see the proposed one more straightforward because all files (models, controllers, views), plain_text/compiled, are first cached and then executed.

niphlod:
@ilvalle: I see. Would you make a PR reverting changes to template and adding your own to compileapp ?
I'm tackling autocompletewidget....

gluon/compileapp.py
```diff
@@ -691,18 +686,18 @@ def run_view_in(environment):
     else:
         filename = pjoin(folder, 'views', view)
         if os.path.exists(path): # compiled views
-            x = view.replace('/', '_')
-            files = ['views_%s.pyc' % x]
+            x = view.replace('/', '.')
+            files = ['views.%s.pyc' % x]
```
ilvalle: compiled views use ```_``` as separator. for example ```views_default_index.html```
https://github.com/web2py/web2py/blob/master/gluon/compileapp.py#L490

PR #1464, opened 2016-09-21
fixes #1309 (and sessions2trash.py, too)

### LINKED ISSUE #1309: expire_sessions.py in admin app
By default admin app in the dist  has nested directory scheme for session files
but current  expire_sessions.py in the cron directory supports the old all-in-one-directory scheme

So it confuses new customers.  
Possible actions:
- remove this example cron task to examples dir
- update it to match nested scheme
- tune admin app for old plain directory session scheme (admin is not for production - has a little sessions)

niphlod:


mdipierro:
Thanks Simone. this is a very important one.

Issue #1441, opened 2016-09-06
pydal tests suite embedded in web2py


niphlod:
a) the way we load tests for pydal... I guess @gi0baro did his best to make them run, but unfortunately cutting corners to have several runs on travis lead to tests being ran BUT not "merged" together in the main result. This means that in order to do a test run on travis for both mysql and postgresql (probably to lower the resulting matrix) we basically gave up tests. Giving that pydal itself doesn't do this, I'd go for restoring the matrix
b)  https://github.com/web2py/pydal/blob/master/tests/sql.py#L1466 groan

I'll open a new issue to discuss how to fix the mess.

niphlod:
guess what "serializers" got tested ? :D gluon's! in fact, no loads_json exists in pydal's helpers.serializers.
guess what tests got run for that test on pydal's tests suite run away from web2py ? None.

niphlod:
once this is merged https://github.com/web2py/pydal/pull/419 , at least tests will be successfull. BTW, I'm astonished that in nearly two weeks noone took a stab at it.

PR #1468, opened 2016-09-23
here here, we can update pydal in the wait for a new release soon

niphlod:
BTW, we should NOW think about https://github.com/web2py/web2py/issues/1441 before hitting the same wall again.


niphlod:
BTW2: this is an "exception" to the normal release process that should wait for a new tagged release from pydal before updating the submodule. Since web2py next release is still "deep in mud" status, at least this makes unittests on travis-ci not failing and completing in a sane amount of time. @gi0baro should be ready to release the september version soon enough.

Issue #1353, opened 2016-06-07
py3 compat tracking


niphlod:
**fix in this before merging #1339**

- [x] gluon/cache.py _a more readable "for key in storage" would be a better pick, and will possible enable leverage future optimization in python_ **py3 incompatibility**
- [ ] contrib/fpdf* _hopefully it's exactly the 1.7.2 as latest stable in its own repo_
- [ ] gluon/contrib/gateways/fcgi.py _those changes IMHO will break something in 2.7 . Given the rise of uwsgi and scgi we should instead rely on the new (and supported) fcgi which in turn depends on flup. A fat warning stating that users relying on fcgi would need flup (which for python/fastcgi hosters is ALWAYS included) would seal the deal_
- [x] gluon/contrib/login_methods/ldap_auth.py _"if type(allowed_groups) != type(list()):" got translated to "if not isinstance(allowed_groups, type(list()))" . Should instead be "if not isinstance(allowed_groups, list)"_
- [ ] gluon/contrib/login_methods/x509_auth.py _won't work with py3, another method is needed_
- [ ] gluon/tools.py : _again, m2crypto is not available for py3_
- [ ] gluon/contrib/shell.py _smells funny, but this would be the 312098312 time we say web shell should not be used. Next release would probably be a REEEEEEALLY good time to drop from the admin app and let the contrib module with a big "won't ever FIX"_
- [ ] gluon/contrib/websocket_messaging.py _won't certainly break 2.7, but I'm not sure it can be considered py3 safe_
- [ ] gluon/globals.py _given the importance of the module, remember to put those #FIXME for py3 in the "let's discuss" recap (or clarify what is left to be decided)_
- [ ] gluon/highlight.py _nothing breaks 2.7 but cgi.escape should be addressed for 3.x_
- [ ] gluon/languages.py _I'd drop iterkeys() alltogether: it doesn't have any improvement over keys() or just the "dry" "for key in dict" syntax_
- [ ] gluon/recfile.py _it has to do with the initial design concept, nothing will break 2.7 but we should really be concerned about relying on auto-closing the stream (py3 will emits warnings about it)_
- [ ] gluon/shell.py as contrib/shell.py: _to me those replacement are not 1:1 . That being said it's on the doctest part which is by far the least utilized part of shell.py_

**maybe valid, but change in behaviour**
- [ ] IS_LENGTH() ?
[...]

PR #1483, opened 2016-09-30
make pymysql usable inside web2py

niphlod:


ilvalle:
Since current contrib/pymysql supports py35, it should no longer necessary to force its installation in ```.travis.yml```. Would you update that file too?

niphlod:
@ilvalle : done. I'll put a relevant note on the issue tracking the new release as we need to figure out and update all NEEDED dependencies before releasing (since it seems the new release will wait for "full" py3 support)

ilvalle:
Web2py core passes almost all unittests using py3, however I wouldn't advertise the new release as py3 compatible because many features are not ready yet. The new release can advertise a _beta_ support or something like that.

Needed dependency are pycrypto (as an alternative to web2py aes) and pg8000. Updating the latter should not be a problem at all.

ilvalle:
In travis build there are few warnings, as for example https://travis-ci.org/web2py/web2py/jobs/164235990#L377
few others are about pending deprecation at line 
```/home/travis/build/web2py/web2py/gluon/packages/dal/pydal/objects.py:2306: PendingDeprecationWarning: generator '__iter__' raised StopIteration```
Those warning regards mainly pydal, not sure are already tracked any where, @gi0baro ?

niphlod:
@ilvalle: in any case, we SHOULD keep track of the "many features that are not ready yet", unless already tracked at #1353 . 
In addition, we exclusively run pydal tests (although the way we do it needs fixing) to test if the shipped drivers are working fine. I'm going to make a PR updating pg8000 and removing the dependency ASAP.
@mdipierro : this is good to merge anyway

PR #1488, opened 2016-10-03
removing pycrypto hard dep

niphlod:
another hard dependency bytes the dust (the last one).

Very very important notes:
- pbkdf2_ctypes.py comes from our very own @michele-comitini : dunno why we lived on a slightly older release
- pbkdf2.py was refreshed from the latest version
- apps using directly gluon.contrib.aes will need to rewrite code (in case) to use pyaes
- @mdipierro : pyaes comes from https://github.com/ricmoo/pyaes . I included license and readme: feel free to drop those in favour of a license recap of all 3rd party
- nowhere in the readme or the book (I noticed just reading the code) mentions that a big speedup on web2py and secure cookies needs pycrypto (or the more versatile pycryptodome) installed...we should:
  - make a note on the book and remember to fetch the dep on the installation scripts for prod environment
  - cook next release binaries with pycrypto included


mdipierro:
Thanks Simone. LGTM but waiting for travis to complete.

PR #1490, opened 2016-10-03
fixes #1485, thanks @abastardi once again

### LINKED ISSUE #1485: response._view_environment missing in scheduler and shell
https://github.com/web2py/web2py/commit/b4acfd0724845633f667a55e593a15576379554d results in a failure to create `response._view_environment` when running scheduler tasks or the shell. Now, `response._view_environment` is only created in `gluon.main.serve_controller`, which only gets called for HTTP requests. A fix, is to create `response._view_environment` in `gluon.shell.env`, after the models have been run (just as in `gluon.main.serve_controller`).

        if import_models:
            try:
                run_models_in(environment)
            except RestrictedError as e:
                sys.stderr.write(e.traceback + '\n')
                sys.exit(1)
        
        response._view_environment = copy.copy(environment)  # <-- ADD THIS LINE

Note, this should actually be an improvement over the previous situation. In the original code, `response._view_environment` was set in `gluon.compileapp.build_environment`, which was run *before* `run_models_in` -- so, the `_view_environment` did not includes objects created in the models. This change rectifies that problem.

niphlod:


ilvalle:
What about adding a scheduler task that uses response.render()? This kind of unittest should cover the use case of the user that found this issue.

niphlod:
you're right, I'll open an issue to keep it tracked.

PR #1501, opened 2016-10-08
Allow cookie to contains several '='

### LINKED ISSUE #1500: webclient : cookie may contains several '='
Which leads to a ValueError: too many values to unpack.

A PR to correct the problem will follow

matclab:
close #1500

### DIFF
--- a/gluon/contrib/webclient.py
+++ b/gluon/contrib/webclient.py
@@ -156,7 +156,7 @@ class WebClient(object):
                 cookie = item[:item.find(';')]
                 pos = cookie.find('=')
                 key = cookie[:pos]
-                value = cookie[pos:]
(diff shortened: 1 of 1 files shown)


ilvalle:
It'd be better to squash the pr into 1 single commit, in addition an unittest would be useful to guarantee former and newer approaches work correctly.

matclab:
Ok, I was not sure about rewriting history of existing PR, but I'll do it.
I'll have a look to adding a unit test to. Thanks for the comment.

matclab:
As i was afraid of, I can not push after squashing the commits… :(
I'll open another PR. I do not know what is the proper accepted way to rewrite history of PR…

PR #1508, opened 2016-10-21
Auth Refactor (WIP)

leonelcamara:
Refactored Auth so you can have different implementations of it's main user functionality (for instance, and implementation not depending on SQLFORM) without breaking backwards compatibility.


mdipierro:
Thank you for this. Give me a couple of days to review it. @niphlod @abastardi comments?

leonelcamara:
I will be adding more tests and methods to the API implementations so the codecov warning should not be a problem soon.
  
My objectives were:
* Backwards compatibility.
* More testability (e.g. you can inject a dummy Auth into the APIs and easily test them and vice versa).
* More flexibility.

Later, more methods can be moved out of Auth if it makes sense, and backwards compatibility will not be a problem.

niphlod:
Don't have time to review it completely but it seems a nice shot. Multiple warnings on the DIDO without csfr are ok, the "real deal" would be to effectively show to end-users why the rewrite makes sense (it actually does it a lot, but without real life examples usually users are not the most embracing people)

abastardi:
Can you explain a bit the design choices of the new architecture and show some examples of how you would use this (both to replicate functionality of the existing system as well as to do something that would currently be difficult)?

A few specific questions:
- Why do we need AuthAPI and DIDO rather than a single class (AuthAPI doesn't implement any methods and simply adds a few attributes in init -- why not just move its init to DIDO and use that as the base class)?
- Why is weakref needed?
- Why call the main class DIDO (I get that it returns dictionaries, but primarily it implements Auth functionality, so it seems odd to focus the name solely on the type of data structures it outputs).
- I haven't examined the code very closely, but could the traditional methods be implemented more easily via composition rather than inheritance? Presumably you could leverage much of the logic of DIDO to implement the Traditional methods rather than completely overwriting the methods.

mdipierro:
I did not do a full review but here are some preliminary comments:
I asked Leonel to look into this. Ideally I would like a low level API for login/register/etc that does not require SQLFORM and a rewrite of the existing ones on top of these low level ones. This code is a great step in the right direction. I like the idea of an interface class (here called AuthAPI) and two classes that extend it (DIDO, Traditional) but I do not like the hierarchy. Naming aside the main problem is that in most programs I do not want to have to chose between DIDO and Traditional. I want often want both. Also, it is not a secret, I want to remove SQLFORM from web2py and therefore I would like an API that can be used in that case. So I think the hierarchy should be AuthAPI (which implements what DIDO does) but should not require Auth, and Auth (which extends AuthAPI, which imports SQLFORM and adds the traditional methods). I would like this interface:

```
 # work as before
auth = Auth(db)
def user(): return dict(form=auth())

# optionally expose an API
auth = Auth(db)
def userapi(): return auth.api()

# only expose an API
auth = AuthAPI(db)
def userapi(): return auth.api()

# where Auth extends AuthAPI
class Auth(AuthAPI):...

# where in both cases I can do
auth.has_....
auth.add_...
etc.
```

Is this way if I want to expose the API I add the userapi() controller action. If I want to replace Auth with AuthAPI() nothing breaks but the user() controller action.

This is very close but I think needs a little more work.

leonelcamara:
@abastardi DIDO is another implementation of some auth methods, it's useful for instance if you don't want to have HTML generation in auth at all (e.g. SPA) and you don't want auth functions to redirect the user. I have AuthAPI and DIDO because in my view DIDO is just one of the many possible implementations of the Auth methods, AuthAPI is not much more than an abstract class that provides the template.  

> Why is weakref needed?  
  
My logic was that I still want to use some functionality in Auth that I believe should stay there as it is generally useful, on the other hand, Auth needs to have an AuthAPI instance too, so this avoids memory leaks by not having a circular reference. Having Auth be an argument to AuthAPI and AuthAPI an argument to Auth also helps with testability as you can use dummy classes.  

> Why call the main class DIDO (I get that it returns dictionaries, but primarily it implements Auth functionality, so it seems odd to focus the name solely on the type of data structures it outputs)  
  
I focused on the data structures because it's already an AuthAPI so I decided to focus on what it differs the most from the other implementation, but I'm certainly open to better name suggestions (naming is hard).  
  
> I haven't examined the code very closely, but could the traditional methods be implemented more easily via composition rather than inheritance? Presumably you could leverage much of the logic of DIDO to implement the Traditional methods rather than completely overwriting the methods.  
  
They could but I would have to make another implementation of SQLFORM based around dicts which was a can of worms I was keen to avoid. DIDO will be a good base to many other AuthAPI implementations I hope, as it doesn't try to do much.

abastardi:
> @abastardi DIDO is another implementation of some auth methods, it's useful for instance if you don't want to have HTML generation in auth at all (e.g. SPA) and you don't want auth functions to redirect the user. I have AuthAPI and DIDO because in my view DIDO is just one of the many possible implementations of the Auth methods, AuthAPI is not much more than an abstract class that provides the template.

Yes, I see what DIDO is doing -- it just seems like AuthAPI is not very useful, as it doesn't actually implement anything. Why not just get rid of AuthAPI and rename DIDO to AuthAPI -- that can then serve as the base class for others to extend? I don't see what an emtpy class gets you (unless maybe we need the additional constraints of an abstract base class, in which case, use an abstract base class).

> My logic was that I still want to use some functionality in Auth that I believe should stay there as it is generally useful, on the other hand, Auth needs to have an AuthAPI instance too, so this avoids memory leaks by not having a circular reference. Having Auth be an argument to AuthAPI and AuthAPI an argument to Auth also helps with testability as you can use dummy classes.

Got it. But do we really need a class structure with circular references? It seems like in an effort to keep the existing Auth class as is, we end up with fairly complicated code. I like Massimo's idea of re-implementing Auth as a subclass of AuthAPI.

leonelcamara:
@abastardi you have some good points there. Indeed DIDO can just be the general base class as long as we strive to always keep it simple so it is indeed useful to inherit from it. I guess I just didn't want to touch Auth too much. I will work more on this.

mdipierro:
@leonelcamara what about the interface I proposed in my comment?

leonelcamara:
@mdipierro yes I will work towards that.

PR #1519, opened 2016-11-01
Allow for firstname and lastname in verify_email message

### LINKED ISSUE #1518: Allow the use of firstname and lastname in auth.messages.verify_email
auth.messages.verify allows only for username to identify people in a name, but username may be the user mail.
It would be nice to allow more personal first and last name.

Other messages may be concerned too.

(I will submit a PR soon).

matclab:
Should fix #1518.

I have not find a simple way to add a test, as there is no existing test for registration.
We could add one by mocking smtp library, but it requires a bit more work to set up.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -3417,7 +3417,9 @@ class Auth(object):
                 link = self.url(
                     self.settings.function, args=('verify_email', key), scheme=True)
                 d = dict(form.vars)
-                d.update(dict(key=key, link=link, username=form.vars[username]))
+                d.update(dict(key=key, link=link, username=form.vars[username],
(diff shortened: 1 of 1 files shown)


matclab:
Hello @mdipierro,

Coming back to this PR, I wonder if I've not been wrongly influenced by the `update(d, …, username=form.vars[username])`.
The line before, there is a `d=dict(form.vars)` which already takes all the fields form `form`.
Moreover, the form fields are call `first_name` and `last_name` (with underscore).

Thus I'm afraid this commit is totally useless and may be reverted… 

I'm so sorry for the noise and the unneeded work.

Issue #1526, opened 2016-11-05
problems in code for reset password flow


michele-comitini:
There are a few worrisome lines of code

never executed:
https://github.com/web2py/web2py/blob/master/gluon/tools.py#L3688

things to avoid in any code: generic exceptions
https://github.com/web2py/web2py/blob/master/gluon/tools.py#L3681
https://github.com/web2py/web2py/blob/master/gluon/tools.py#L3684

a catchall exception:
https://github.com/web2py/web2py/blob/master/gluon/tools.py#L3685

a mysterious test parameter which could potentially leak some security info:
https://github.com/web2py/web2py/blob/master/gluon/tools.py#L3687

leonelcamara:
3687-3688 seems like whoever was developing it forgot a line with a test redirect. Sort of a redirect version of print debugging. I'll just remove it in my auth refactor. The cactchall exception is not as bad in my opinion as it's working like a kind of "goto error handling"  for the various kinds of errors one can have with the key, I'm inclined to leave it for now.

michele-comitini:
@leonelcamara 
IMHO to understand what goes on at least one should 
raise Exception('meaningful description')

mdipierro:
can this be closed?

leonelcamara:
@mdipierro sort of, we still have the catchall exception in confirm_registration which would probably be a good idea to refactor.

PR #1527, opened 2016-11-05
Auth refactor

leonelcamara:
Extracted many methods into a base class for more generic auth mechanisms.

Partially addresses #1526
Includes a solution for IS_LOWER and IS_UPPER validator problems I mentioned in #1353

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-sudo: false
+sudo: required
 
 cache: pip
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-sudo: required
+sudo: false
 
 cache: pip
 
--- a/gluon/authapi.py
+++ b/gluon/authapi.py
@@ -698,6 +698,10 @@ class AuthAPI(object):
                 value = user[key]
                 if callable(value) or key == self.settings.password_field:
                     delattr(user, key)
+        current.session.auth = Storage(user=user,
+                                       last_visit=current.request.now,
+                                       expiration=self.settings.expiration,
+                                       hmac_key=web2py_uuid())
         return user
 
     def login_user(self, user):
@@ -707,10 +711,6 @@ class AuthAPI(object):
         user = self._update_session_user(user)
         if self.settings.renew_session_onlogin:
             current.session.renew(clear_session=not self.settings.keep_session_onlogin)
-        current.session.auth = Storage(user=user,
-                                       last_visit=current.request.now,
-                                       expiration=self.settings.expiration,
-                                       hmac_key=web2py_uuid())
         self.user = user
         self.update_groups()
 
--- a/gluon/authapi.py
+++ b/gluon/authapi.py
@@ -698,12 +698,13 @@ class AuthAPI(object):
                 value = user[key]
                 if callable(value) or key == self.settings.password_field:
                     delattr(user, key)
+        return user
 
     def login_user(self, user):
         """
         Logins the `user = db.auth_user(id)`
         """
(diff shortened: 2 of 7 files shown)


leonelcamara:
@mdipierro @niphlod @abastardi comments?

niphlod:
sorry for the "not-so-active" review process but I'm swamped. A quick look (no tests, just general flow) gives me no shivers, so LGTM. We're not seeing a new stable for a while anyway, there's time to fix things.

mdipierro:
I am revising this in the next few days.

mdipierro:
This looks really good. I am going to take another couple of days to test it and then merge it.

niphlod:
@leonelcamara : no sudo in travis, pleeeease !!!!!

mdipierro:
I agree. But how do we solve the problem of the postgres installation? without that tests do not pass.

niphlod:
solving an issue breaking how the suite runs isn't a fix, it's a band-aid ready to come off.

niphlod:
see https://www.traviscistatus.com/incidents/ftk8v7ms1ks6 . they've been having problems with the container infrastructure.

leonelcamara:
@niphlod sudo does make travis run slower because it can't use a container, but how does it break anything? All it does is make travis use the traditional build infrastructure we always used. 
  
This is just going around what is arguably a travis bug with their new docker containers as we aren't really using sudo anywhere I can see in our travis.yml, a bandaid indeed, hopefully we can remove it soon.

niphlod:
Shouldn't change the output but I don't get the "pressure".
And btw "always" is relative, we use the container env since it has been
available (2 years maybe?)

Il 21 nov 2016 6:58 PM, "Leonel Câmara" <notifications@github.com> ha
scritto:

leonelcamara:
@niphlod ok, you're obviously right on this one, I just changed it back, and things do seem to be fixed already.

niphlod:
not obviously right, just faithful in travis which until now served us greatly. ^_^

leonelcamara:
@mdipierro is there anything missing or that you need improved?

nextghost:
I do have a feature request. I'd like an `Auth` method that will take a `dict` of user record values, register a new user and do all the background registration work, including sending a verification e-mail and requesting admin approval.

I need it for custom public registration form so using `register_bare()` would be a security disaster waiting to happen (if I screw up the check for existing records before calling it, it'll happily overwrite them; and it doesn't send verification e-mails anyway).

The new method should also have parameters for forcing/overriding e-mail verification and admin approval:
- `True` - do it regardless of global settings
- `False` - don't do it
- `None` (default) - use global settings

mdipierro:
This looks great to me. If nobody objects I will merge this tomorrow.

leonelcamara:
@nextghost that's actually one those things that really easy to do by subclassing AuthAPI

nextghost:
@leonelcamara It's also one of those things that would be hard to maintain because it's duplication of existing upstream code that may change without notice.

leonelcamara:
@nextghost I would be up for having a version, inheriting from this one that sent mails too in the trunk if it's a common usage. In fact I used to have emailing in AuthAPI, I decided to remove it because some people may want to inherit from AuthAPI and not have emails at all.  
  
Once in the trunk AuthAPI will not change, apart from bug fixing, without notice as web2py does try very hard to maintain backwards compatibility.

nextghost:
@mdipierro If you're keeping this PR open only because of me, you can go ahead and merge it. I'll open a new pull request for my idea later.

mdipierro:
I promise I will merge this very soon. this topic is very dear to me so I want to make sure I am 100% fine with it and I think I am.

PR #1529, opened 2016-11-10
Mobile (iOS) fix for dismissing flashes

vinyldarkscratch:
iOS devices don’t like listening to clicks on most objects.  They
typically prefer a and button objects.  This fix replaces the
#closeflash span with a link to “javascript:null;” instead (while also inheriting
text-decoration and color styling), so that mobile (iOS) devices will
allow you to close the flash.


vinyldarkscratch:
Had some strange Git sync issues, and I'm not certain why my tests are failing -- nothing I modified should cause those particular test fails.  I recreated this as pull request #1531, and fixed a few of my errors.

PR #1531, opened 2016-11-11
Mobile (iOS) flash dismissal fix

vinyldarkscratch:
iOS devices don’t like listening to clicks on most objects. They
typically prefer a and button objects. This fix replaces the
#closeflash span with a link to “javascript:null;” instead (while also inheriting
text-decoration and color styling), so that mobile (iOS) devices will
allow you to close the flash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 *.orig
 Thumbs.db
 .DS_Store
+*.DS_Store
 index.yaml
 routes.py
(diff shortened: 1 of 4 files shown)


vinyldarkscratch:
Seems the check fails occurring with Travis CI are due to [pydal](https://travis-ci.org/web2py/pydal/jobs/174852504), not due to my code.

PR #1532, opened 2016-11-11
Fix T.M() not adding to language files

vinyldarkscratch:


mdipierro:
This fails Travis. Could you check?

vinyldarkscratch:
Hey @mdipierro, the reason for Travis failing is because PyDAL happens to be failing tests -- seems that it's an error with postgres, where a warning shows waiting for the user to press Enter.  Codecov, I'm not too sure why that's failing, will have a look soon!

mdipierro:
oK. looking into it.. thanks.

PR #1533, opened 2016-11-17
implemented base support for CASv3

gchiesa:
This commit extend the base CAS support also to v3 protocol. The only assumption is that the url for cas login for 'servicevalidate' and 'proxyvalidate' for V3 of protocol is like the following: 

```
- v3:
provider/default/user/cas/p3/{serviceValidate, proxyValidate}
- v2:
provider/default/user/cas/{serviceValidate, proxyValidate}
```


mdipierro:
Thank you! very useful.

PR #1539, opened 2016-12-01
when mkdir of missing app folders, handle broken links properly

smorrison:
The code for creating missing application folders doesn't take symlinks into account.

If a web2py app is configured to have symlinks for some system folders (in my case `uploads` exists on a larger volume), and the link is broken because the target dir was removed, then the code in try_mkdir won't work.

`if not os.path.exists(path)` will return True when path is a broken symlink, but `os.mkdir(path)` will fail because the file exists (as a symlink). 

The fix is to identify broken links and mkdir the target instead of the link.

### DIFF
--- a/gluon/admin.py
+++ b/gluon/admin.py
@@ -438,7 +438,11 @@ def add_path_first(path):
 def try_mkdir(path):
     if not os.path.exists(path):
         try:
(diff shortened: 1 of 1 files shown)


mdipierro:
Thanks good fix.

PR #1540, opened 2016-12-03
Replaced ipaddr with ipaddress.py (backported from py3)

ilvalle:


mdipierro:
Can you please send more details about the license of IPaddress? our policy is to include only libs licensed under BSD or equivalent.

ilvalle:
@mdipierro it's released under the _PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2_
see https://github.com/phihag/ipaddress/blob/master/LICENSE

PR #1544, opened 2016-12-09
response.layout

zvolsky:
What about this?
Together with setting (inside model, controller)
 response.layout="layout_stupid.html"
 response.layout="layout_w3.html"

it could help to switch standard/mobile, standard/experimental, ... designs.

Good or bad idea?


mdipierro:
I think this is bet handled by editing the file. This breaks bytecode compilation of views. In a next version of web2py we should only use responsive layouts.

zvolsky:
>> bytecode compilation

Ok, please Reject this PR.

mdipierro:
LOL. Thanks anyway.

PR #1558, opened 2017-01-13
Fix router functions check

abastardi:
Consider the following router:

    routers=dict(
        BASE=dict(
            default_controller='default',
            default_function='index',
                functions=['index', 'func1', 'comp1']
        )
    )

Now include the following component in a view:

    {{=LOAD('default', 'comp1.load', ajax=True)}}

This component will make an Ajax request to `/comp1.load`. The router must then determine whether `"comp1.load"` is a URL arg for the `/default/index` function, or whether it is actually a function itself (to be routed to `/default/comp.load`). To make this determination, it simply checks whether `"comp.load"` is in the `router['BASE']['functions']` list. This check fails, so the request gets routed to `/default/index/comp.load` rather than the correct `/default/comp.load`.

Currently, the `functions` list in the router only works properly if it includes all permutations of functions and extensions that the app will use. In the above example, it would have to be:

                functions=['index', 'func1', 'comp1.load']

This pull request strips the extension before checking the functions list, so a request to "comp.load" would correctly match "comp" in the functions list.

Perhaps there is some rationale for the current behavior, but in that case, the rewrite code that maps outgoing URLs must be changed. Given the above example, `URL('default', 'comp1.load')` would generate `/comp1.load`, which will then be routed to `/default/index/comp1.load`. It should instead generate `/default/comp1.load` -- otherwise, there is no way to generate a URL for the `/default/comp1.load` route.

### DIFF
--- a/gluon/rewrite.py
+++ b/gluon/rewrite.py
@@ -1040,7 +1040,7 @@ class MapUrlIn(object):
         else:
             default_function = self.router.default_function  # str or None
         default_function = self.domain_function or default_function
-        if not arg0 or functions and arg0 not in functions:
+        if not arg0 or functions and arg0.split('.')[0] not in functions:
             self.function = default_function or ""
             self.pop_arg_if(arg0 and self.function == arg0)
         else:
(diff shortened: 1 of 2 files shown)


leonelcamara:
Nice catch! Would be perfect with a test 👍

abastardi:
OK, I added some tests, which now pass with the fix.

Issue #1561, opened 2017-01-21
Provide a way to write "service" functions that accept *args/**kwargs


BrenBarn:
Currently, if you write a "service" function using the decorators like `@service.json`, etc., it can't usefull accept varargs.  This is because the `universal_caller` function in gluon.tools only passes arguments that explicitly correspond to named arguments in the service function's signature.  (This in turn is likely because `universal_caller` is working based on things like `co_argcount`, which doesn't provide a way to see whether a function accepts varargs.)  See mailing list discussion [here](https://groups.google.com/forum/#!topic/web2py/Xlrobpwv5Aw).

It would be nice to be able to write a service function that accepts arbitrary `*args` and/or `**kwargs`, passed as part of the URL path or as URL parameters, respectively.

Since `universal_caller` is a standalone function, this would probably require creating a new analogue of the Service class that rewrites the various `serve_*` methods to use a different function instead.  (Modifying `Service` directly could change the behavior of existing code.)  The simplest alternative is just:

```
def new_universal_caller(f, *a, **b):
    f(*a, **b)
```

This is basically what `request.restful` does.

If we do want to maintain the convention of passing only accepted arguments, we could use something like this:

```
def new_universal_caller(f, *a, **b):  
    argspec = inspect.getargspec(f)

    requiredArgs = argspec.args if argspec.defaults is None else argspec.args[0:-len(argspec.defaults)] 

    # if it accepts *args, we pass all positional arguments, otherwise only the number it accepts
    args = a if argspec.varargs is not None else a[:len(argspec.args)]
    
    # if it accepts **kwargs, we pass all keyword arguments, otherwise only those it accepts
    kwargs = b if argspec.keywords is not None else {argName: b[argName] for argName in argspec.args if argName in b}
       
    # it's an error to pass an argument both positionally and by keyword
    if any(argName in argspec.args[:len(args)] for argName in kwargs):
        raise HTTP(404, "Argument passed positionally and by keyword")
    # if we didn't get enough positional arguments, any missing ones must have been passed by keyword
    if any(argName not in kwargs for argName in requiredArgs[len(args):]):
        raise HTTP(404, "Not enough arguments passed")
    # otherwise, ok
    else:
        return f(*args, **kwargs)
```

`inspect.getargspec` provides a way to tell if the function accepts varargs or not.

[...]

Issue #1560, opened 2017-01-18
Some Service types lumped together incorrectly


BrenBarn:
In gluon.tools, the Service class defines a number of attributes run_procedures, json_procedures, etc., for storing different types of service functions.  However, some of the decorator methods store the decorated functions in an incorrect place.  For instance:

```
    def csv(self, f):
        """
        Example:
            Use as::

                service = Service()
                @service.csv
                def myfunction(a, b):
                    return a + b
                def call():
                    return service()

            Then call it with::

                wget http://..../app/default/call/csv/myfunction?a=3&b=4

        """
        self.run_procedures[f.__name__] = f
        return f
```

You can see at the end it is putting it in `run_procedures`, but this is the `csv` decorator, so it should be putting it in `csv_procedures`.  The `xml` decorator likewise stores the function it gets under `run_procedures` rather than `xml_procedures`.  In fact, the `csv_procedures` and `xml_procedures` attributes are never used at all.

More worrisome, both `serve_csv` and `serve_xml` actually look in `run_procedures` rather than `csv_procedures` or `xml_procedures` to see if the passed service is valid.  This means that if you define a function with `@service.csv`, you will be able to access it with a URL like `.../call/run/my_function`, and vice versa.  Effectively there is no distinction between "run", CSV, and xml services.  This can potentially lead to strange errors; you can for instance define a `@service.run` function that doesn't return data in a CSV format, which will cause an error if you try to access it via `../call/csv` since the handler is unable to parse it as CSV.

Looking at git blame, it appears the error has been present for several years (maybe since Service was added?), but apparently hasn't been noticed.

ilvalle:
@BrenBarn https://github.com/web2py/web2py/pull/1566 has been merged, would you close this issue?

PR #1566, opened 2017-01-28
Add FlexibleService, which allows @service-style methods with varargs

BrenBarn:
This addresses #1561 .

I don't know what your testing procedure is for things like this.  I see that the current test_tools.py just has TestService as "todo" so this would be in line with that.  I've done some simple tests by creating a sample app and calling the service endpoints and it works.  In any case, this change only adds a new class, so it shouldn't affect any existing behavior.

The PR defines a new FlexibleService class that inherits from the existing Service.  It overrides that various `serve_whatever` methods so that instead of using `universal_caller`, they use a method `call_service_function` provided by FlexibleService.  The default implementation of this method just directly calls the function, passing whatever arguments were given to the "call" controller function.  This means the service function can accept `*args` and/or `**kwargs`.  This also means that if the service function doesn't accept the passed arguments, an exception will be raised, rather than returning an "Object does not exist" 404 as Service currently does.  Using a method for `call_service_function` instead of a module-level function means that it's possible to write new classes that inherit from FlexibleService and add more sophisticated argument checking if they want to.

The new FlexibleService class also addresses bug #1560 by fixing the errors in the CSV and XML methods so that they store and find the procedures in the right place.  This does not, however, fix the bug in Service itself that stores these procedures in the wrong place.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -6053,6 +6053,396 @@ class Service(object):
     def error(self):
         raise HTTP(404, "Object does not exist")
 
+class FlexibleService(Service):
+    # __init__ inherited
+    
+    # we override csv and xml to fix a bug that caused these procedures to be stored under run_procedures instead of csv_procedures or xml_procedures
+    # other service decorators (run, json, etc.) are inherited
+
+    def csv(self, f):
+        """
+        Example:
+            Use as::
+
+                service = Service()
+                @service.csv
+                def myfunction(a, b):
+                    return a + b
+                def call():
+                    return service()
+
+            Then call it with::
+
+                wget http://..../app/default/call/csv/myfunction?a=3&b=4
+
+        """
+        self.csv_procedures[f.__name__] = f
+        return f
+
+    # fix bug where stuff was put in run_procedures instead of xml_procedures
+    def xml(self, f):
+        """
+        Example:
+            Use as::
+
+                service = Service()
+                @service.xml
+                def myfunction(a, b):
+                    return a + b
+                def call():
+                    return service()
+
+            Then call it with::
+
+                wget http://..../app/default/call/xml/myfunction?a=3&b=4
+
+        """
+        self.xml_procedures[f.__name__] = f
+        return f
+
+    def serve_run(self, args=None):
+        request = current.request
+        if not args:
+            args = request.args
+        if args and args[0] in self.run_procedures:
+            return str(self.call_service_function(self.run_procedures[args[0]],
+                                        *args[1:], **dict(request.vars)))
+        self.error()
+
+    def serve_csv(self, args=None):
+        request = current.request
+        response = current.response
+        response.headers['Content-Type'] = 'text/x-csv'
+        if not args:
+            args = request.args
+
+        def none_exception(value):
+            if isinstance(value, unicodeT):
+                return value.encode('utf8')
+            if hasattr(value, 'isoformat'):
+                return value.isoformat()[:19].replace('T', ' ')
+            if value is None:
+                return '<NULL>'
+            return value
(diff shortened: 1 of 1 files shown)


abastardi:
I don't think we should simply add a whole new class to the API (particularly given that it simply duplicates most of the original code, which is not DRY). Rather, we should fix the current `Service` class by changing `universal_caller` to accept arbitrary arguments (though moving it to a method of the `Service` class is probably a good idea, to facilitate subclassing). If there is some good reason to retain the current behavior as the default, then we should simply add an argument to the `Service.__init__` method to enable the new behavior as an option -- something like:

    class Service(object):

        def __init__(self, environment=None, allow_packing=False):
            ...
            self.allow_packing = allow_packing

        def universal_caller(self, f, *args, **kwargs):
            if self.allow_packing:
                [new code]
            else:
                [old code]

And then change later calls from `universal_caller(...)` to `self.universal_caller(...)`.

mdipierro:
I agree with Anthony. Let me know if anybody is willing to work on it, else I will do this.

BrenBarn:
Note that that can potentially change the behavior of existing uses of `Service`, although the kinds of usages that it would break are probably uncommon.  I agree it would be preferable but I hesitated to do that because of how it might change existing behavior.

mdipierro:
I do not think it would be a problem, because, as Anthony suggests there would be an optional argument to the Service() constructor (allow_arg_packing) and if false the current behavior stays. If True it will follow the new behavior you propose. Looks like you already have some code to test this on so would be great if you could do it.

BrenBarn:
Do we consider `universal_caller` as a public function from `gluon.tools` that people might be using on its own?  Or is it okay to move it inside `Service` as a method?

abastardi:
I don't think `universal_caller` is considered part of the public API, so should be safe to move it inside the `Service` class (if there is any concern, we could always leave it where it is but still create a new method in the class for actual use by the class).

The only question is whether `allow_arg_packing` should default to `False` (identical to the current behavior) or `True` (more convenient, though if this breaks backward compatibility for anyone, they would have to explicitly set the arg to get the old behavior). I lean toward `True` because it's not clear this would really be a compatibility issue, and if so, probably very rarely, and with an easy fix.

@mdipierro, thoughts?

BrenBarn:
I changed the code so it modifies Service instead of adding FlexibleService.  I left `universal_caller` in place, but just added the `call_service_function` method that calls `universal_caller` if we want the old behavior, or doesn't call it if we want the new behavior.

Service now has an argument to decide between old and new behavior.  I named this argument `check_args`, which I think more accurately reflects its meaning: the new behavior is not just "allow passing args/kwargs"; it's "just pass everything, without checking what arguments the service function accepts at all".  The default is False, i.e., don't check args, which means this defaults to the new behavior, in line with what you suggested.

BrenBarn:
Not sure what is up with the failing tests, they don't seem to involve Service at all. . .

niphlod:
@BrenBarn not your fault. @mdipierro checked in some code that shouldn't have been committed in master.

BrenBarn:
Okay, well, let me know if/when I should rebase and update the PR.

mdipierro:
Hello @BrenBarn, I merged it and apologies it took me so long. But can you please write a line in the book to describe the new syntax?

BrenBarn:
Sorry, I've been away from this for a bit.  I'm not sure what really needs to be added to the docs.  It's rather that the existing documentation didn't document the limitations of the previous implementation.  It was never documented that extra arguments would be silently ignored, or that the service function couldn't accept args/kwargs.  We could add a note saying explicitly that you now can do that?

PR #1578, opened 2017-02-26
minor cleanup removed unused class 'navbar-ex1-collapse' see https://…

ndegroot:
layout.html had an unused class picked up from an (now) old example on bootstrap site. See https://github.com/twbs/bootstrap/issues/10948.
Nothing serious, wasn't referenced anywhere. Just fluff.


mdipierro:
thanks and sorry for the late response

PR #1589, opened 2017-03-12
Small fixes to AuthJWT

kristjanvalur:
- The time stamp was being generated in a needlessly complicated fashion.  In addition, it was inconsistent in its use of UTC; which could have led to errors in non-utc locales. Simplified this.

- the internal api clearly allows for the altered payload to be returned from the alter_payload function, or the callable self.additional_payload member.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -1346,7 +1346,7 @@ class AuthJWT(object):
             exp=expires,
             hmac_key=web2py_uuid()
         )
-        self.alter_payload(orig_payload)
+        orig_payload = self.alter_payload(orig_payload)
         return orig_payload
 
     def alter_payload(self, payload):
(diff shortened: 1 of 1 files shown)


niphlod:
hum. gmtime() returns a struct, not an int.... are you sure the code is working ?

kristjanvalur:
Sorry, please strike that.  I'll revoke this.

kristjanvalur:
I'll fix it, there still was an inconsistency I was trying to address.
Btw, since I have you here, what was your intention with the return value from alter_payload?

niphlod:
technically none. the need was just to have a function that would alter the payload, not something that returns a specific value

kristjanvalur:
Right.  Ok, in that case IMHO, it would be cleaner not to assign and return a return value in the callback function, that then gets ignored.
If you dont mind, I‘ll clean that up if I manage to get this patch working (I‘m an old mercurial hand, trying to grapple with GIT for the first time)
K

niphlod:
NP.

PR #1591, opened 2017-03-14
Some fixes on ldap_auth

cccaballero:


BuhtigithuB:
Think to squash your changes if possible...

But good catch for logging/logger

BuhtigithuB:
But are you sure ? As logging.SOMETHING still good as it get import it can
be use... logger seems web2py specific...



On Tue, Mar 14, 2017 at 2:10 PM, Richard Vézina <ml.richard.vezina@gmail.com
> wrote:

> Think to squash your changes if possible...
>
> But good catch for logging/logger
>
>
>
> On Tue, Mar 14, 2017 at 1:15 PM, Carlos Cesar Caballero Díaz <

cccaballero:
logger is the logging instance of the module (you can look at lines 180-188), in a few code lines was used logging instead of logger and the messages where not showing.

About squashing the changes, I am not sure what I need to do.

BuhtigithuB:
One way would be to revert your change the last change and --amend the
previous commit or you can stash, but I guess rebase would be the best
option :
https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

Before trying anything copy your repository beside the other and try there
before you master squashing...

mdipierro:
Thanks and sorry for the late reply

PR #1593, opened 2017-03-15
Improve french translation and close #1592

### LINKED ISSUE #1592: LOAD(ajax=True) is missing a translation
When I call `{{=LOAD('default','page.load', ajax=True)}}` in a view, it displays the text "loading..." while the component is being loaded. On my version of web2py (Version 2.14.6-stable+timestamp.2016.05.10.00.21.47, on OS X), the string 'loading...' is not translated into different languages, e.g. if I visit the page with a browser set to default to french.

BuhtigithuB:


BuhtigithuB:
Commit seems to fail testing

PR #1604, opened 2017-04-06
prevent is_empty from stripping whitespaces, close #1570

### LINKED ISSUE #1570: IS_NOT_EMPTY and IS_EMPTY_OR should not strip whitespace
When checking values for being empty, the IS_NOT_EMPTY and IS_EMTPY_OR validators first strip whitespace to ensure the value isn't just whitespace. The problem is that these validators then return the stripped version of the value instead of the original value. This means that values that are not just whitespace but include some whitespace will be transformed before being inserted in the database. For example, " hello " will be transformed to "hello". This behavior is unexpected, as the purpose of these validators is not to remove whitespace from otherwise non-empty strings.

It may be appropriate to strip whitespace in order to determine whether a string should be considered empty (and we probably have to continue that to maintain backward compatibility), but I think returned the stripped value instead of the original value should be considered a bug. If the developer wants to strip whitespace from submitted values, that should be handled explicitly via another validator or some other method.

See https://groups.google.com/forum/#!topic/web2py/_SfUxePTaN0.

ilvalle:
updated few tests.


mdipierro:
Technically this is a change of behavior but I see this should be considered the correct behavior.

PR #1606, opened 2017-04-10
Added on_succes and custom options for web2py ajax function

cccaballero:
Added posibility to append an "add" hook and rewrite default jquery ajax options in the web2py ajax function.

Example:

    ajax{
        "url",
        ['name'],
        'target',
        {
            add: function(){
                console.log("ok!!!");
            },
            error: function(){
                console.log("error!!!");
            }
        }
    }

### DIFF
--- a/applications/admin/static/js/web2py.js
+++ b/applications/admin/static/js/web2py.js
@@ -38,7 +38,7 @@
             if (value > 0) $('#' + id).hide().fadeIn('slow');
             else $('#' + id).show().fadeOut('slow');
         },
-        ajax: function (u, s, t) {
+        ajax: function (u, s, t, options) {
             /*simple ajax function*/
             var query = '';
             if (typeof s == 'string') {
@@ -59,7 +59,9 @@
                     query = pcs.join('&');
                 }
             }
-            $.ajax({
+
+            // default options for jquery ajax function
+            var ajax_options = {
                 type: 'POST',
                 url: u,
                 data: query,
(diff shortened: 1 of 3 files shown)


leonelcamara:
"on_success" and "fail"? It should either be "on_success" and "on_fail", or simply success and fail, I realize you did it so it would not clash with the success the ajax function already uses so why not call it "done" and "fail". Note that fail is not working as it should go in "error" on the ajax options.  
  
Why not something like this:

```
var ajax_call = $.ajax(ajax_options).done(put_the_regular_success_function_here_and_remove_success_from_ajax_options);
if (options.done) {
    ajax_call.done(options.done);
}
if (options.fail) {
    ajax_call.fail(options.fail)
}
// We could add a alway here
```

Alternatively "success" can be an array of functions, so you could simply add the options.on_success funcion there.

cccaballero:
Yes, there is an error in my example, it should be "error" instead of "fail" (this happen when you code in many tecnologies at same time), because the idea is to put any of the jquery ajax options and an extra "on_suceess" option (I will rename it to "done" because I think is better to)
I will edit the main comment and update the request right now.

cccaballero:
But why use done(), fail() or complete() if the ajax options accepts arrays of functions?
Right now the use of done() is discouraged against the use of the Promise syntax, an in the future will be adopted async and await, but the "success", "error" and "complete" options will remain the same using in the background the recommended implementation.

leonelcamara:
@cccaballero agreed, that seems to be the best option.

mdipierro:
I am taking this but I think it is best to use jQuery.ajax or axios.js instead of the built-in ajax function. It will never be on pair to the others.

PR #1607, opened 2017-04-12
appadmin can use ccache with cache.ram = cache.disk = cache.memcache

sugizo:
ref : https://groups.google.com/forum/#!topic/web2py/ycOvyOBPCMQ


ilvalle:
Looks go to me

PR #1610, opened 2017-04-20
handle reference fields for keyed tables

geomapdev:
added code to support reference fields to keyed tables and reference fields that don't have a format defined in the referenced table.


ilvalle:
Would you add an unittest for such enhancement? see examples in gluon/tests

geomapdev:
I may need some help on that part, it looks like a unittest hasn't been created yet for default validators:

https://github.com/web2py/web2py/blob/master/gluon/tests/test_dal.py#L45

ilvalle:
@geomapdev for an example have a look at https://github.com/web2py/web2py/blob/master/gluon/tests/test_validators.py#L204  in addition you should check that a default validator exists for those field types defined in dal.py

mdipierro:
looking at this. probably a go but need to check more carefully.

gluon/dal.py
```diff
@@ -51,11 +51,29 @@ def _default_validators(db, field):
...
-            if field.unique:
-                requires._and = validators.IS_NOT_IN_DB(db, field)
-            if field.tablename == field_type[10:]:
-                return validators.IS_EMPTY_OR(requires)
-            return requires
+        else:
+            requires = validators.IS_IN_DB(db, referenced._id)
+        if field.unique:
+            requires._and = validators.IS_NOT_IN_DB(db, field)
+        if field.tablename == field_type[10:]:
```
geomapdev: should be "if not field.notnull:"?

geomapdev:
A few other things I noticed, it looks like like this if statement should have an else:
https://github.com/web2py/web2py/blob/master/gluon/dal.py#L51

and lines 54-58 should not be indented.  (similar to list:reference below it)

Let me know if you'd like me to add those changes.

ilvalle:
@geomapdev PR looks good, however few lines are simply repeated. to be more DRY I'd just ask refactoring the PR in order to wrap together the common code there is to handle `reference` field type with & without keyed tables ( and the same for the `list:reference` field type).

Regarding the unittest see the comment above

mdipierro:
@geomapdev thanks for the PR. Please let us know if you can make the requested changes.

geomapdev:
Sorry, things have been really busy for me lately.  I will try to tackle it this week.

geomapdev:
Updated code to be more DRY, added unittest under test_validators.py.  

Let me know if there's anything else that needs to be addressed.

geomapdev:
Also added a unittest for "list:reference table" field type.

Attempted to add a unittest for "list:reference table.field" and a keyed table, but there is a validator error when trying to insert into the table on this line: 
https://github.com/geomapdev/web2py/blob/dal-enhancements/gluon/tests/test_validators.py#L325

Looks like there would be changes required in validators.py to handle "list:reference table.field" on a keyed table, but all the other reference tests are successful.

mdipierro:
Thank you for this work1

PR #1617, opened 2017-04-27
fix open file in py3

ilvalle:


mdipierro:
Really we should make sure we use "with open_file() as" everywhere. Anyway this is an improvement.

PR #1632, opened 2017-05-15
Fixes #1628 missing signature list

### LINKED ISSUE #1628: NameError: global name 'signature_list' is not defined
```python
  File "/Users/abhishesh/Dropbox/WORK/web2py/gluon/restricted.py", line 216, in restricted
    exec(ccode, environment)
  File "/Users/abhishesh/Dropbox/WORK/web2py/applications/artemis/models/db.py", line 231, in <module>
    auth.define_tables(username=False, signature=False, enable_tokens=True)
  File "/Users/abhishesh/Dropbox/WORK/web2py/gluon/tools.py", line 2238, in define_tables
    settings.table_token_name, []) + signature_list
NameError: global name 'signature_list' is not defined
```

leonelcamara:


gluon/tools.py
```diff
@@ -2214,7 +2214,19 @@ def define_tables(self, username=None, signature=None, enable_tokens=False,
...
+        if signature is True:
+            signature_list = [self.signature]
+        elif not signature:
+            signature_list = []
+        elif isinstance(signature, Table):
+            signature_list = [signature]
+        else:
+            signature_list = signature
+
+        super(Auth, self).define_tables(username, signature_list, migrate, fake_migrate)
```
mdipierro: why is the code above necessary? authapi.define_tables already has the same exact code

leonelcamara: The problem is that Auth.define_tables also needs to have signature_list available if it wants to define more tables than AuthAPI. Would you prefer that AuthAPI.define_tables sets an auth instance variable with the signature list (e.g. self._table_signature_list)?

mdipierro:
yes. that would be better. thanks

leonelcamara:
I submitted a better version with your suggestions.

mdipierro:
Thanks!

Issue #1636, opened 2017-05-31
sqlhtml.py throws exception when there is a left in query


timrichardson:
Line 2332 on sqlthml.py has a problem for me. 
This is the context: 

The grid causing this:

```
grid = SQLFORM.grid(db_vciadmin.patient_entity_link,
                        left = db_vciadmin.appt_reminder_status.on(db_vciadmin.patient_entity_link.unified_id == db_vciadmin.appt_reminder_status.clinic_patient), ...
```

snippet surrounding line 2332 
```
 tablenames = db._adapter.tables(dbset.query)  #tablenames is a dictionary
        if left is not None:
            if not isinstance(left, (list, tuple)):
                left = [left]
            for join in left:
                tablenames += db._adapter.tables(join)  #this is line 2332
```
The last line in that snippet throws an exception. 
`<type 'exceptions.TypeError'> unsupported operand type(s) for +=: 'dict' and 'dict'
`

` db._adapter.tables(join) `evaluates to
```
{'appt_reminder_status': <Table appt_reminder_status (id, appt_summary_id, email_id, appt_confirmed, callback_id, sms_id, appt_date, email_sent_when, email_reply_when, sms_sent_when, second_sms_sent_when, sms_reply_when, location, entity, patient_name, patient_phone, patient_email, clinic_patient, email_link_visit, appt_startdatetime, inactive_row, callback_url, sms_process_complete, email_process_complete, confirmation_method, notes, sms_fail_message_sent, email_fail_message_sent, do_not_contact, sms_reply_text)>,
 'patient_entity_link': <Table patient_entity_link (id, unified_id)>}
```
so we are trying to do += on two dicts

timrichardson:
Note that the problem may be that 
`gluon/packages/dal/pydal/adapters/base.py` tables(...) is returning a dict.

abastardi:
> Note that the problem may be that
gluon/packages/dal/pydal/adapters/base.py tables(...) is returning a dict.

Yes, `db._adapter.tables()` was changed to return a dict instead of a list at [this commit](https://github.com/web2py/pydal/commit/6b5e1b860e03030e265490ad01b0ba6b388eb3bb).

timrichardson:
this is the fix. I'll do a pull request when I get a moment.

```
from pydal.helpers.methods import smart_query, bar_encode, _repr_ref, merge_tablemaps # new import
.
.
.

   tablenames = db._adapter.tables(dbset.query) #is now a dict
        if left is not None:
[...]

PR #1637, opened 2017-06-01
fix to #1636: sqlgrid left= (left outer join) fails because db._adapt…

timrichardson:
…er.tables() now returns dict (previously a list)


mdipierro:
Good catch! thanks.

PR #1643, opened 2017-06-06
Fixes #1598

### LINKED ISSUE #1598: URL having character | in fragment part, does not pass IS_URL validator
`IS_URL()` validator doesn't accepts following url and similar google group urls. 
Example: https://groups.google.com/forum/#!searchin/web2py/url$20validation$20problem|sort:relevance/web2py/69KmsrYAF44/Sgk_hA_6TlcJ

Character | should be allowed in fragment part of the url

leonelcamara:
Also added a few more tests for IS_URL


ilvalle:
@leonelcamara is it visible to introduce `urlparse`  (from gluon._compat import urlparse) in place of `url_split_regex`? By doing that a future integration of py3 will be easier

leonelcamara:
@ilvalle I'll make an attempt.

PR #1644, opened 2017-06-07
Changed URL validation to use urlparse instead of regex for spliting …

### LINKED ISSUE #1598: URL having character | in fragment part, does not pass IS_URL validator

leonelcamara:
…the URL

Enabled test_is_url in Python 3 since it is now passing
This might be one of the last fixes to #1353
Fixes #1598

### DIFF
--- a/gluon/tests/__init__.py
+++ b/gluon/tests/__init__.py
@@ -24,7 +24,7 @@ from .test_web import *
 from .test_sqlhtml import *
 from .test_scheduler import *
 from .test_cron import *
+from .test_is_url import *
 
 if sys.version[:3] == '2.7':
-    from .test_is_url import *
     from .test_old_doctests import *
--- a/gluon/tests/test_is_url.py
+++ b/gluon/tests/test_is_url.py
@@ -586,81 +586,81 @@ class TestUnicode(unittest.TestCase):
                # disables prepending the scheme in the return value
 
     def testUnicodeToAsciiUrl(self):
-        self.assertEquals(unicode_to_ascii_authority(u'www.Alliancefran\xe7aise.nu'), 'www.xn--alliancefranaise-npb.nu')
-        self.assertEquals(
+        self.assertEqual(unicode_to_ascii_authority(u'www.Alliancefran\xe7aise.nu'), 'www.xn--alliancefranaise-npb.nu')
+        self.assertEqual(
             unicode_to_ascii_authority(u'www.benn.ca'), 'www.benn.ca')
         self.assertRaises(UnicodeError, unicode_to_ascii_authority,
                           u'\u4e2d' * 1000)  # label is too long
 
     def testValidUrls(self):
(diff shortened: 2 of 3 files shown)


leonelcamara:
Everyone please review this as much as possible, I only wanted to fix #1598 but sneaky @ilvalle suggested I could use urlparse in my previous pull request #1643. The changes ended up being much more extensive and harder than I thought. Ohh well, at least this is one of the most tested files in web2py.

ilvalle:
The PR looks go to me.
A minor question, why is the try/except no longer necessary in https://github.com/web2py/web2py/pull/1644/files#diff-1e4f774e944ce88607f6bc3debe6f0b7L1473 ?

Regarding https://github.com/web2py/web2py/issues/1353 few checkboxes have been fixed, @niphlod would you update it?

niphlod:
Going to. You know the list of ones I forgot excluding this, of course?

Il 07 giu 2017 18:08, "ilvalle" <notifications@github.com> ha scritto:

leonelcamara:
@ilvalle That try except was actually quite a weird situation because test_is_url.py in https://github.com/web2py/web2py/pull/1644/files#diff-99e8ab468a65ee093a207f013a521837R592

In a line that I have not changed, actually tests that an exception is raised and try except was preventing the exception from being raised in python 3 causing the test to fail.  In python 2 it actually seemed to work before because it would raise an UnicodeError inside the except clause whereas that did not happen with python 3.

ilvalle:
@leonelcamara thank for the explanation
@niphlod in #1353 you can check, `IS_LENGTH`, `IS_IP`, `port web2pyDialog`, `execute the entire test suite on 3.5`. Regaring `cache.disk` I'd ear @leonelcamara option

leonelcamara:
cache disk tests seem to pass pretty well on python 3 so it seems to be fine now.

niphlod:
#1353 updated.

ilvalle:
@niphlod you can check `execute the entire test suite on 3.5`, `test_old_doctests` is the only one missing and it isn't worth to port it.

niphlod:
@ilvalle : are you saying that utf8.py and markmin don't need to be tested ? doctests cover html.py, utf8.py and markmin2hhtml. html.py has a corresponding unittest suite, the other 2 no. 

BTW: here is kinda offtopic, better discussing #1353 directly

mdipierro:
Can you please resolve the conflicts with master and resubmit?

leonelcamara:
@mdipierro done, ironically the conflict arose because of my other pull request being merged which changed the same import line.

PR #1647, opened 2017-06-11
Update html.py

auden-rovellequartz:
adding in helpers for the following HTML tags:

<article></article>
<footer></footer>
<header></header>

### DIFF
--- a/gluon/html.py
+++ b/gluon/html.py
@@ -40,6 +40,7 @@ entitydefs.setdefault('apos', u"'".encode('utf-8'))
 
 __all__ = [
     'A',
+    'ARTICLE',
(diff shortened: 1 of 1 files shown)


mdipierro:
Auden. I would rather not take this. The reason is that we have lots of requests for adding new helpers and we do not want to keep polluting the namespace. But there is a simple solution. In your code, one of the models, you can do:
```
ARTICLE, HEADER, FOOTER = TAG.article, TAG.header, TAG.footer 
```
and define the headers you need.

auden-rovellequartz:
Ok - - - and thanks for the tip ☺

PR #1666, opened 2017-06-21
Enable Ajax file uploads

abastardi:
These changes to web2py.js allow file uploads via Ajax in browsers that support the `FormData` API. Older browsers will continue to fall back to the jQuery `.serialize()` method.


abastardi:
Note, this affects Ajax components that contain forms (via the `trap_form` function) but does not affect usage of the `ajax()` function.

PR #1692, opened 2017-07-21
Fix #1691: Passing tables to SQLFORM.factory() fails

### LINKED ISSUE #1691: Passing tables to SQLFORM.factory() fails
Example:
`form = SQLFORM.factory(idb.catering, idb.catering_diet)`

File "/home/www-data/internal-apps/gluon/sqlhtml.py", line 1922, in factory
    return SQLFORM(DAL(None).define_table(table_name, *[field.clone() for field in fields]),
AttributeError: 'Table' object has no attribute 'clone'

Since .clone() was added to the list comprehension, it tries to clone tables as well, rather than passing them on to define_table() as is.

timnyborg:
Still clone fields, in factory(), but leave tables untouched so they get passed to define_table correctly

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -1919,13 +1919,8 @@ class SQLFORM(FORM):
         if 'table_name' in attributes:
             del attributes['table_name']
 
-        fields_with_clones = []
-        for field in fields:
-            if isinstance(field, Field):
-                fields_with_clones.append(field.clone())
-            else:
-                # We have a table, so pass it along
-                fields_with_clones.append(field)                
+        # Clone fields, while passing tables straight through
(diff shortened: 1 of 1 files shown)


leonelcamara:
Wouldn't it be simpler to turn this:  

    *[field.clone() for field in fields]

into:

    *[f.clone() if isinstance(f, Field) else f for f in fields]

It should be faster too.

timnyborg:
That's what I had, but feared it might prove difficult to read

leonelcamara:
@timnyborg yes they are faster. I think they are quite readable in this kind of "filter" situations too.

leonelcamara:
Looks great to me. Maybe close and reopen the pull request to trigger a travis rebuild, I see no reason why it should fail.

PR #1722, opened 2017-08-06
mobilize is back

### LINKED ISSUE #1721: @Mobilize support missing in 2.15.2?
After upgrading to 2.15.2 views using **mobilize** statement throw an exception.
Before (2.14.6) same views worked fine

I can reproduce this using a minimal example 

from gluon.contrib.user_agent_parser import mobilize
@mobilize
def index():
   return dict()

throws an error
<type 'exceptions.ImportError'> cannot import name mobilize

File "/home/www-data/web2py/gluon/restricted.py", line 219, in restricted
exec(ccode, environment)
File "/home/www-data/web2py/applications/myapp/controllers/stats.py", line 3, in <module>    from gluon.contrib.user_agent_parser import mobilize

ImportError: cannot import name mobilize

Any ideas?

leonelcamara:
Fixes #1721

### DIFF
--- a/gluon/contrib/user_agent_parser.py
+++ b/gluon/contrib/user_agent_parser.py
@@ -678,6 +678,9 @@ def simple_detect(agent):
 class mobilize(object):  
     """
     Decorator for controller functions so they use different views for mobile devices.
+
+    WARNING: If you update httpagentparser make sure to leave mobilize for
+    backwards compatibility.
     """
     def __init__(self, func):
         self.func = func
--- a/gluon/contrib/user_agent_parser.py
+++ b/gluon/contrib/user_agent_parser.py
@@ -673,3 +673,20 @@ def simple_detect(agent):
     if os_version:
         os = " ".join((os, os_version))
     return os, browser
+
+
+class mobilize(object):  
+    """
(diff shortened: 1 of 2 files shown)


mdipierro:
Did you check this? The problem as I see it is that user_agent.is_mobile is no longer defined and always return None.

leonelcamara:
@mdipierro you are right. I did not.  

The problem seems to be that we used to have is_mobile and is_tablet in our detectors and we lost them on the update to httpagentparser 1.7.8.  
    
I think the whole concept of checking if something is_mobile or is_tablet is outdated so we should probably just remove this feature. making sure to also remove it in cache.action.  

Otherwise this pull request need to be put on hold until I add is_mobile and is_tablet to the proper detectors.

mdipierro:
@leonelcamara I agree.

leonelcamara:
@mdipierro this is working now. I followed a different approach to try and not pollute httpagentparser as much as possible so I "detect" mobile and tablet in Request now.

So, if you decide to keep mobilize, you can merge it. Otherwise you can close it.

mdipierro:
Nice! Thanks you.

PR #1726, opened 2017-08-07
auth.has_membership(cached=True) check membership in auth.user_groups

BuhtigithuB:


leonelcamara:
This looks good to me, I just don't understand why all the if/elses

```python
        if cached:
            if user_id and group_id in self.user_groups:
                r = True
            else:
                r = False
        else:
            membership = self.table_membership()
            if group_id and user_id and self.db((membership.user_id == user_id) &
                                                (membership.group_id == group_id)).select():
                r = True
            else:
                r = False
```

```python
        if cached:
            if user_id and group_id in self.user_groups:
                r = True
            else:
                r = False
        else:
            membership = self.table_membership()
            if group_id and user_id and self.db((membership.user_id == user_id) &
                                                (membership.group_id == group_id)).select():
                r = True
            else:
                r = False
```

Couldn't this be just:

```python
        if cached:
            r = user_id and group_id and group_id in self.user_groups:
        else:
            membership = self.table_membership()
            r = user_id and group_id and self.db((membership.user_id == user_id) &
                                                (membership.group_id == group_id)).select()
```

Note that this still makes a query to the database for the case where the role is given and not the id. One could argue that perhaps you should cache the group names in the session as well to avoid the queries completely.

BuhtigithuB:
Yes, it could be clearer like that... I only intent to make sure I didn't break anything...

And yes, it still made a query because id_group()...

My (mostly Simone is the author) original snippet was : 

```python
def has_membership(id_role):
    return True if (auth.user_id and id_role in auth.user_groups.values()) or \
                   (auth.user_id and id_role in auth.user_groups) else False
```

So, I maybe better do :

```python
        if not user_id and self.user:
            user_id = self.user.id
        if cached:
            id_role = group_id or role
            if user_id and id_role in self.user_groups:
                r = True
            else:
                r = False
        else:
            group_id = group_id or self.id_group(role)
            try:
                group_id = int(group_id)
            except:
                group_id = self.id_group(group_id)  # interpret group_id as a role
            membership = self.table_membership()
            if group_id and user_id and self.db((membership.user_id == user_id) &
                                                (membership.group_id == group_id)).select():
                r = True
            else:
                r = False
        self.log_event(self.messages['has_membership_log'],
                       dict(user_id=user_id, group_id=group_id, check=r))
```

What do you think??
­­

leonelcamara:
It's fine then, we can take care of caching the role later.  
  
I would just add a commit adding this functionality to Auth.requires_membership.

BuhtigithuB:
I mean more like this : 

```python 
        if not user_id and self.user:
            user_id = self.user.id
        if cached:
            id_role = group_id or role
            if (user_id and id_role in self.user_groups.values()) or (user_id and id_role in self.user_groups):
                print user_id and id_role in self.user_groups
                r = True
            else:
                print user_id and id_role in self.user_groups
                r = False
        else:
            group_id = group_id or self.id_group(role)
            try:
                group_id = int(group_id)
            except:
                group_id = self.id_group(group_id)  # interpret group_id as a role
            membership = self.table_membership()
            if group_id and user_id and self.db((membership.user_id == user_id) &
                                                (membership.group_id == group_id)).select():
                r = True
            else:
                r = False
        self.log_event(self.messages['has_membership_log'],
                       dict(user_id=user_id, group_id=group_id, check=r))
        return r
```
I made a mistake in the cached if forgeting that I need an OR : 
if (user_id and id_role in self.user_groups.values()) or (user_id and id_role in self.user_groups):
In this use case...

I check, and this : 
(user_id and id_role in self.user_groups.values()) or (user_id and id_role in self.user_groups)

Would return a True/False, will make sure the older code do it to and remove a couple of if/else...

I will look at Auth.requires_membership...

Wait before merging I will push a better commit version...

BuhtigithuB:
Can't remove the if/else in the older code as it returns a db rows object :(

mdipierro:
ready to be merged?

BuhtigithuB:
No, wait... I will finish this change ASAP and tell you here when it ready...

BuhtigithuB:
Ok I push the final version, tests pass locally, if everything goes write with travis and appveyor, you can merge!!

:)

BuhtigithuB:
Ready! Please add the documentation PR in the book, except if you want to keep it experimental for sometimes... PR is pending so you have options...

PR #1728, opened 2017-08-08
Don't minify css or js files in other applications

### LINKED ISSUE #1673: can't use response.optimize_js in models/db.py, previous version can work
in models/db.py
response.optimize_js = 'concat,minify,inline'

  File "/Users/MacBookPro/site/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "<string>", line 36, in <module>
  File "/Users/MacBookPro/site/web2py/gluon/globals.py", line 501, in include_files
    time_expire)
  File "/Users/MacBookPro/site/web2py/gluon/cache.py", line 247, in __call__
    value = f()
  File "/Users/MacBookPro/site/web2py/gluon/globals.py", line 496, in call_minify
    self.optimize_js)
  File "/Users/MacBookPro/site/web2py/gluon/contrib/minify/minify.py", line 101, in minify
    contents = read_binary_file(abs_filename)
  File "/Users/MacBookPro/site/web2py/gluon/contrib/minify/minify.py", line 19, in read_binary_file
    f = open(filename, 'rb')
IOError: [Errno 2] No such file or directory: '/Users/MacBookPro/site/web2py/applications/welcome/static/min.js'

leonelcamara:
Fixes #1673

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -476,8 +476,9 @@ class Response(Storage):
         files = []
         ext_files = []
         has_js = has_css = False
+        app = current.request.application
         for item in self.files:
-            if isinstance(item, (list, tuple)):
+            if isinstance(item, (list, tuple)) or not item.startswith('/' + app):
                 ext_files.append(item)
                 continue
             if extensions and not item.rpartition('.')[2] in extensions:


leonelcamara:
This fixes it but is breaking tests because of two reasons, 1 the tests need to be slightly changed because response now cares about which application the request was made to, and 2 it changes the order of the included files.

In fact 2 seems to be a problem that already existed. include_files cannot change the order of the included files as that may break the final javascript and css result. If you look at the tests in test_include_files you see that it actually expects that the order will be changed. This should **not** happen.

PR #1761, opened 2017-09-13
Update groups on edit profile

josedesoto:
When profile is updated `self._update_session_user(user)` set session.user_groups to None.  After the update self.update_groups() should be executed to keep session consistent.


mdipierro:
I am happy to take your fix but you used a tab instead of spaces for indent. Can you fix it?

PR #1785, opened 2017-09-22
Enhances/fixes validators/is_empty() for Python 3 compatibility, fix #1764

### LINKED ISSUE #1764: Bug in gluon/validators.py/is_empty() in Python 3
I was digging around trying to save an edit/sqlform with an upload field and it was failing with "Invalid image" error message. My upload field is using "requires=IS_EMPTY_OR(IS_IMAGE())" validator.
Since the field already have an image and I just want to update others fields it does not make sense to fire this error. Debugging a bit I found that upload field value (from input) is sent empty (b'') and the is_empty() function (called from IS_EMPTY_OR()) is returning "NOT EMPTY" cause only value '' and [] are treated as empty (not b''). It worked fine when I changed the bold line in is_empty() function to:

<pre>
def is_empty(value, empty_regex=None):
    _value = value
    """test empty field"""
    if isinstance(value, (str, unicodeT)):
        value = value.strip()
        if empty_regex is not None and empty_regex.match(value):
            value = ''
    <b>if value is None or value == '' or value == b'' or value == []:</b>
        return (_value, True)
    return (_value, False)
</pre>

Thanks in advance.

tiagobar:
Fix #1764

### DIFF
--- a/gluon/tests/test_validators.py
+++ b/gluon/tests/test_validators.py
@@ -444,6 +444,8 @@ class TestValidators(unittest.TestCase):
         self.assertEqual(rtn, (None, 'Enter a value'))
         rtn = IS_NOT_EMPTY()('')
         self.assertEqual(rtn, ('', 'Enter a value'))
+        rtn = IS_NOT_EMPTY()(b'')
+        self.assertEqual(rtn, (b'', 'Enter a value'))
         rtn = IS_NOT_EMPTY()('  ')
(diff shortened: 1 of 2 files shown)


ilvalle:
@tiagobar would you mind to cover this corner case with an unittest in https://github.com/web2py/web2py/blob/master/gluon/tests/test_validators.py. This will guarantee that future changes to web2py will not introduce this issue again.

Issue #1795, opened 2017-11-02
Internal middleware returns incorrect status code


vitkarpenko:
Example from http://web2py.com/books/default/chapter/29/04/the-core#Internal-middleware returns 500 status code on latest web2py version (while still converting 'hello world' to 'HELLO WORLD' correctly).

vitkarpenko:
I've create a pull request for this problem: #1796

PR #1796, opened 2017-11-02
gluon.http expects status code as integer while LazyWSGI generates str

vitkarpenko:
Fix for #1795


ilvalle:
@vitkarpenko can you close #1795 ?

PR #1803, opened 2017-11-13
Fix #1715

### LINKED ISSUE #1715: parameter of ajax put request is empty in Python 3
I tested the latest version 2.15.2 and I've got a problem when accessing ajax post parameters with web2py under Python 3.

The client sends a PUT request using jquery like this:
`$.ajax('<url>', {
    data: JSON.stringify({
        outputFormat: 'pdf'
    }),
    type: "PUT", contentType: "application/json",
    success: function(data) {
        alert('success');
    },
    error: function(jqXHR, textStatus, errorThrown) {
        alert('error');
    }
});`

in the controller I access the parameter
`output_format = request.vars.outputFormat`

when using Python 2.7 the variable output_format is set to 'pdf' as expected. When using Python 3.5 the variable is None.

leonelcamara:


ilvalle:
what do you think about checking the python version before calling to_native?

leonelcamara:
I don't think it's worth it. In 3.6 internally the decoding will have to happen anyway and in 2.7 nothing will be done inside to_native.

PR #1825, opened 2017-12-12
python3 compatibility for woth sorted-cmp as well as unicode

roaldatwork:
To fix the languages editor in admin application when running python 3

### DIFF
--- a/gluon/_compat.py
+++ b/gluon/_compat.py
@@ -73,6 +73,10 @@ if PY2:
             return obj
         return obj.encode(charset, errors)
 
+    def u(x):
+        "safe alternative for the python 2 `unicode()` function, which doesn't exist in python 3"
+        return unicode('x', 'utf-8')
+
 
 else:
     import pickle
@@ -132,6 +136,10 @@ else:
         if obj is None or isinstance(obj, str):
             return obj
         return obj.decode(charset, errors)
+
+    def u(x):
+        "safe replacement for the python 2 `unicode()` function"
+        return x
 
 
 def with_metaclass(meta, *bases):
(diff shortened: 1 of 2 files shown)


roaldatwork:
I discovered that the translate functionality of the admin application doesn't work in python 3.
I found the gluon._compat.py file, and implemented the u() function according to [this topic](http://python3porting.com/noconv.html#more-bytes-strings-and-unicode)

ilvalle:
isn't to_native enough? In addition, see comment on the code

gluon/_compat.py
```diff
@@ -73,6 +73,10 @@ def to_native(obj, charset='utf8', errors='strict'):
             return obj
         return obj.encode(charset, errors)
 
+    def u(x):
+        "safe alternative for the python 2 `unicode()` function, which doesn't exist in python 3"
+        return unicode('x', 'utf-8')
```
ilvalle: the `x` parameter is not used at all

roaldatwork: hmm fine by me as well. didn't know that one :)

roaldatwork: and whoops - i almost create a bug

roaldatwork: thats what we have peer reviews for - glad you catched it

roaldatwork:
Sorry, I was to hasty to get the my fix in there. Fixed it now

gluon/_compat.py
ilvalle: I suggest closing this PR and opening a new one that uses to_native without the `u()` method definition

roaldatwork: I followed up your advise. Thanks for the feedback

roaldatwork:
I understand that - I'm too busy now, but will do that later today

PR #1834, opened 2018-01-07
Added the web2py_disable_garbage_collect option to avoid garbage collect

ilvalle:
Added an option to disable garbage collect


mdipierro:
Hello Paolo, what is the reason for this patch?

ilvalle:
To let user to disable garbage collection in certain conditions (for example during benchmarks)

PR #1839, opened 2018-01-25
Python 3 compatibility issues

maxcrystal:


ilvalle:
@maxcrystal mind that this PR changes pydal version. This is not correct. It's better to close this PR and open a new one with the single fix to gluon/shell.py

PR #1843, opened 2018-02-01
Allow choosing a SAML entityid

timnyborg:
Allows you to pick an IDP's entityid when using federation XML data or an MDQ with many entries.  
Pysaml2's MDQ metadata is loaded lazily, after you provide a key, so it's necessary to pass the entityid as a key.


ilvalle:
@timnyborg thanks for submitting a PR. Is it python3 compatible?

timnyborg:
Yep, nothing py3-unfriendly

PR #1852, opened 2018-02-08
fix py3 languages decode 

ilvalle:


mdipierro:
Notice this change does not pass some tests. strange because it does not change py2 behavior.

PR #1860, opened 2018-02-12
patched flash function to prevent XSS

blackthorne:
if an attacker controllable input reaches web2py's notification system (flash() function), a XSS becomes possible. Using .text() should prevent such issues since injected content will be escaped and not rendered as Javascript.

### DIFF
--- a/applications/admin/static/js/web2py.js
+++ b/applications/admin/static/js/web2py.js
@@ -617,7 +617,7 @@
         flash: function (message, status) {
             var flash = $('.w2p_flash');
             web2py.hide_flash();
-            flash.html(message).addClass(status);
+            flash.text(message).addClass(status);
             if (flash.html()) flash.append('<span id="closeflash"> &times; </span>')[animateIn]();
         },
         hide_flash: function () {
--- a/applications/examples/static/js/web2py.js
+++ b/applications/examples/static/js/web2py.js
@@ -617,7 +617,7 @@
         flash: function (message, status) {
             var flash = $('.w2p_flash');
             web2py.hide_flash();
-            flash.html(message).addClass(status);
+            flash.text(message).addClass(status);
             if (flash.html()) flash.append('<span id="closeflash"> &times; </span>')[animateIn]();
         },
         hide_flash: function () {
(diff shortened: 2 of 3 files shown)


leonelcamara:
Humm but the message can have HTML. I don't see how flash can have anything from an untrusted source unless you put it there, if you put it there it's your responsibility to escape it. This is like saying response.js is a security vulnerability.  
  
I think a good middleground may be to use $.parseHTML with keepscripts false. But changing this to text is not acceptable as it breaks currently working applications for a very very small security increase for idiots.

blackthorne:
By untrusted source I don't necessarily mean unauthenticated. It's easy to think of an application with users of different roles that can create objects that will show up in those notifications and therefore would allow to run JS in the context of a different user perhaps with different level of access. It certainly is possible to escape these things on their various entry points but it seems a lot more reliable / safe to do it in the place where they will be brought up. 

I think with a notification system like this expectations are that it only needs text so it's easy to neglect the need to sanitize things prior to submission. With response.js it's a much more generic channel which explicitly adds JS to a response, so no assumptions are made towards security.

My assertions and subsequent pull request comes from real applications that I have assessed with this problem and given the impact the security increase is not that negligible. Finally, this should not break any user applications since their web2py.js remains untouched even following a web2py update.

leonelcamara:
Can you give me an example of how do you have this vulnerability, I mean without the programmer explicitly doing something like: 

    response.flash = XML(user_provided_input, sanitize=False)

> Finally, this should not break any user applications since their web2py.js remains untouched even following a web2py update.  
  
Humm that's true. I drop that point.

blackthorne:
I have seen apps that wire a dedicated level of Python's logging system to flash notifications and today I saw one that used websockets and flash for general alerts and messaging. I can't disclose this code (does not belong to me) but if you can think of this p.o.c. code:
https://github.com/web2py/web2py/blob/master/gluon/contrib/websocket_messaging.py

where on line 57 you would change it into something like:
 `var callback=function(e){$.web2py.flash(e.data)};`

now if your input is `<script>prompt(1)</script>` you will get a popup.

leonelcamara:
Ohh I didn't consider websockets, good catch.

blackthorne:
any updates on this? is it to be closed?

leonelcamara:
@blackthorne no don't close it, this will either be merged or we will have to find another alternative.

@mdipierro ping

Issue #1866, opened 2018-02-19
Cache size on appadmin: guppy ImportError


dartg:
The `ccache` function inside `appadmin.py` (http://127.0.0.1:8000/yourapp/appadmin/ccache) uses [guppy](https://pypi.python.org/pypi/guppy/) (if available) to show the memory used by `cache.ram`. But once installed (and if there's something stored in the RAM cache) it will fail with the following traceback:

> <type 'exceptions.ImportError'> (ImportError("Cannot import module 'applications.[appname].modules.guppy'",), <traceback object at 0x0907F968>)

The exception is started at [`appadmin.py:455`](https://github.com/web2py/web2py/blob/82cf2eb44db907a3ca20c29ee8116f95ef77bb5a/applications/welcome/controllers/appadmin.py#L455):

`                ram['bytes'] += hp.iso(value[1]).size`

And specifically raised at `guppy/etc/Glue.py:250` where `__import__` is used. Seems like this somewhat conflicts with the web2py import system.

masia5:
Guppy is no longer maintained. You could replace it with a recursive version of `sys.getsizeof ` see https://docs.python.org/3/library/sys.html#sys.getsizeof . Would you try to post a PR?

dartg:
Even the recursive version of `sys.getsizeof` will return wrong results for pyDAL records and user-defined classes. Maybe [Pympler](https://pypi.python.org/pypi/Pympler) could be used, which is newer and pure-Python.

ilvalle:
@dartg https://github.com/web2py/web2py/pull/1869 has been merged. you can close the issue

dartg:
Great!

PR #1869, opened 2018-02-20
Replace the guppy library with Pympler

dartg:
Since guppy is no longer mantained (#1866) I replaced it with [Pympler](https://pypi.python.org/pypi/Pympler), which is newer, pure-Python and provide better results!


mdipierro:
Thanks. Did not know about this.

PR #1878, opened 2018-03-13
Explain command switched specific to sessions2trash.p

BuhtigithuB:


BuhtigithuB:
It tooks me sometimes to figure why processes were spawned without be ended...

:D

PR #1898, opened 2018-04-17
fix editing mistakes

richboss:


leonelcamara:
Minor correction to your corrections, you corrected "formsts" to "forms" when it should be "formats".

PR #1942, opened 2018-06-07
Use request_reset_password +  PDF and TSV exports

jvanbraekel:
Update /user/request_reset_password to user/retrieve_password and linked settings option in user.html.
According to https://web2py.readthedocs.io/en/latest/_modules/gluon/tools.html#Auth 'request_reset_password' should not even be exposed anymore. (but this is still the case).


jvanbraekel:
New change : fix markmin2latex import for pdf file generation.

PR #1948, opened 2018-06-12
removed merging get vars and post vars with the same name as list

yamandu:
it was breaking some code that excpected a single variable, like SQLFORM.grid
is this really necessary?

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -276,10 +276,6 @@ class Request(Storage):
         for key, value in iteritems(self.post_vars):
             if key not in self._vars:
                 self._vars[key] = value
-            else:
-                if not isinstance(self._vars[key], list):
-                    self._vars[key] = [self._vars[key]]
-                self._vars[key] += value if isinstance(value, list) else [value]
 
     @property
     def get_vars(self):


leonelcamara:
That's necessary because you can actually send a list of values that way.
e.g.
http://127.0.0.1:8000/default/index?productType=Clothing&productType=Bags

The fact that it joins post and get vars with the same key in a list is just a side effect.

yamandu:
Removing this portion of code, it continues to parse these repeat vars as a list with POST and GET.
Before this is executed the list are already parsed.
`    def parse_get_vars(self):
        """Takes the QUERY_STRING and unpacks it to get_vars
        """
        query_string = self.env.get('query_string', '')
        dget = urlparse.parse_qs(query_string, keep_blank_values=1)
        # Ref: https://docs.python.org/2/library/cgi.html#cgi.parse_qs
        get_vars = self._get_vars = Storage(dget)
        for (key, value) in iteritems(get_vars):
            if isinstance(value, list) and len(value) == 1:
                get_vars[key] = value[0]
`

So the effect is really merge POST and GET vars with the same name as list.
My question is if this exact behavior is necessary

leonelcamara:
The point is that you can have a get and post var with the same name, and web2py lets you access all the vars in request.vars, your way simply discards the second value. I don't think that's right, I think it makes you have hidden bugs.  
  
If the grid has a problem (what's the problem by the way?) with it then probably it's the grid that has to be changed to make sure it doesn't happen.  
  
Which by the way, is an example of exactly this, simply discarding the value would hide the bug in the grid.

mdipierro:
Sorry Leonel is right

PR #1954, opened 2018-06-15
Catch KeyError in download function

gonguinguen:
This change is necessary to avoid an error ticket when you delete an upload field from a table (and also from the model definition). In those cases, requests to old URLs will throw a KeyError. With this fix, those requests return 404 as expected.

Also, catching AttributeError is still necessary for the situation when the table is removed entirely. If an old request come in, that sentence will throw an AttributeError, that will be catched properly, and request will return 404.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -667,7 +667,7 @@ class Response(Storage):
         (t, f) = (items.group('table'), items.group('field'))
         try:
             field = db[t][f]
-        except AttributeError:
+        except (AttributeError, KeyError):
             raise HTTP(404)
         try:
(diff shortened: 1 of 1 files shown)


gonguinguen:
I don't know why the codecov tests failed. Actually, I don't know exactly what it checks. I've seen the details but couldn't figure out. In anycase, let me know if I can help.

leonelcamara:
@gonguinguen they just check the code coverage of the automated tests (lines of code executed by the test/total lines of code), since you added lines of code without adding new tests it's normal that you get a lower coverage hence codecov complains about it. It's not necessarily something to worry about, although, of course, we aim at increasing coverage on the long term.

Issue #1993, opened 2018-08-25
file anyserver.py is not python3.5 ready


JusticeN:
Hi everyone,
when I start web2py with the command below it works as expected
`sudo python anyserver.py -s gunicorn -i 127.0.0.1 -p 80 -l`

but running the same command with python 3 throw me some syntax error
`sudo python3 anyserver.py -s gunicorn -i 127.0.0.1 -p 80 -l`

the Print statement is still without parentheses. 
does anyone is using it in production?

thanks in advance

PR #2002, opened 2018-09-05
changing the anyserver.py to make it run with py3

JusticeN:
actually fixing the issue #1993 i have created.
i tested it successfuly with the command python3 anyserver.py -s tornado -i 127.0.0.1 -p 80 -l

### DIFF
--- a/anyserver.py
+++ b/anyserver.py
@@ -212,16 +212,16 @@ def mongrel2_handler(application, conn, debug=False):
 
     while True:
         if debug:
-            print "WAITING FOR REQUEST"
+            print("WAITING FOR REQUEST")
 
         # receive a request
         req = conn.recv()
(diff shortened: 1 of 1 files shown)


tiagobar:
Is this Python 2 compatible?
I think it should be necessary to add the future before line 12:

from __future__ import print_function

leonelcamara:
@tiagobar print('hello') works fine on python 2.7 for me

tiagobar:
> @tiagobar print('hello') works fine on python 2.7 for me

Ok, you're right. print as a function call like python 3 works on 2.7, I didn't know this! However it isn't python 3 compatible. The parameters from python3 print like "print('hello', end='')" don't works. But it's ok how it is used in this case. Thanks for your reply.

PR #2008, opened 2018-09-06
Add SameSite support

leonelcamara:
This adds an extra layer of security to web2py to help prevent CSRF and is a feature that nowadays is supported by most browsers.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -1075,6 +1075,8 @@ class Session(Storage):
             scookies['HttpOnly'] = True
         if self._secure:
             scookies['secure'] = True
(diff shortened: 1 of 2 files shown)


mdipierro:
Love this. Should we default self._same_site=True when session.secure()? I am tempted to say it should always be set to True.

leonelcamara:
I'm inclined to agree that almost everyone wants at least Lax.

mdipierro:
OK. Let's default to lax.

leonelcamara:
@mdipierro done! Made SameSite=Lax the default

Issue #2007, opened 2018-09-06
SQLFORM.factory Returns Date field as string (str) instead of date (in form.vars) as in 2.14.06


icodk:
After upgrade to 2.17.1 (from 2.14.06)
SQLFORM.factory  date filed is returned as 'str' type  field instead of 'date'

As a bonus issue, the session.flash does not work either (see below).
 
**To Reproduce**
Run the following function and check the form.vars.today value printed

function:
```

def form_test():
    form = SQLFORM.factory(
        Field('today', 'date', label='today', writable=True, default=request.now.date()),
        buttons=[   TAG.button(T('Roll back'), _type="submit", _class="btn btn-primary")
                ,   TAG.button(T('Cancel'), _type="button",_onClick="parent.location='%s' ;return false" % URL('default', 'form_test'))
                 ], table_name='test_tbl',
    )
    title=T("Form test")

    if form.process(session=None, formname='test').accepted:
        print form.vars.today,type(form.vars.today) #this line prints 'str' as the field type 
        session.flash = T('Message that not flashing') #this message does not flash
        redirect(URL('default', 'form_test'))
    elif form.errors:
        response.flash = 'form has errors'
    else:
        response.flash = T('Select items to refund')
    return locals()

```
The view (form_test.html):

```
{{left_sidebar_enabled=right_sidebar_enabled=True}}
{{extend 'layout.html'}}
{{block header}}
 {{end}}
{{block left_sidebar}}
 {{end}}
 {{block right_sidebar}}
 {{end}}
 <div class="alert alert-success" role="alert"><h3 align=center>
         {{=title}}
              </h3>
 </div>
    <div class="well">
	<div class="row">
        <div class="col-md-9">
        <h4>{{=T('Fields with writable=False shows None')}}</h4>
        {{=form.custom.begin}}
        <table id="mytable" class="table table-bordred table-striped">
                   <thead>
                   </thead>
        
    <tbody>
      <tr><td>{{=form.custom.widget.tag_product_name}}</td>
    </tr>
   <tr>
   <td align="right">{{=form.custom.widget.today}}</td>
    </tbody>
    </table> 
            {{=form.custom.submit}}
            {{=form.custom.end}}
</div>
</div>
</div>

```

**Expected behavior**
Return field type as  'date'  (as in web2py version 2.14.06)

leonelcamara:
That's weird. If you add a requires=IS_DATE() to the field does it fixes the problem?

icodk:
[...]

PR #2019, opened 2018-09-22
Testcase for #2007

hrother:


mdipierro:
hello hrother. I like th PR but it does not pass the travisci tests:
```
FAIL: test_factory_applies_default_validators (gluon.tests.test_sqlhtml.TestSQLFORM)
----------------------------------------------------------------------
  File "gluon/tests/test_sqlhtml.py", line 333, in test_factory_applies_default_validators
    self.assertIsInstance(factory_form.vars.a_date, datetime.date)
AssertionError: '2018-09-14' is not an instance of <type 'datetime.date'>
```
Can you fix it so that it does?

leonelcamara:
@mdipierro I think the problem is that this is a test for the #2007 fix which wasn't merged yet

mdipierro:
My bad. Merging.

PR #2026, opened 2018-09-27
Python 3 compatibility + fixed link for cookbook

nicozanf:


nicozanf:
After approval (if it's right, obviously) someone must update the web2py's web site - also the download page is obsolete

PR #2032, opened 2018-10-07
Create generic.csv

rayluo:
[The book said "web2py does not define a generic.csv file"](http://web2py.com/books/default/chapter/29/10/services#CSV). I think we can still provide one default implementation to suit the most basic need. The application developers can always implement their own view file.

PS: When/if this PR would be accepted, please consider also changing that sentence in the book.

On a side note, I was the author of the generic.load too. It feels good to contribute to web2py again. :-)

### DIFF
--- a/applications/welcome/views/generic.csv
+++ b/applications/welcome/views/generic.csv
@@ -1,0 +1,17 @@
+{{"""Usage:
+
+  def controller():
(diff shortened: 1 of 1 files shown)


leonelcamara:
Can you close and reopen the pull request just to make travis run again? You got a stupid apt install error.

PR #2040, opened 2018-10-18
fix plugin.html view to work with python3

d3im:
Admin app has error in this view under python3 - it could be seen when there is any plugin installed.


d3im:
please merge ASAP, this fix doesn't break anything and manual correction on every update/new installation is really not what I like to do... ;-)

PR #2045, opened 2018-10-20
Fixes python3 traceback errors 

veiko99:


jicho:
@veiko99 when you have an error in a view file, the complete file isn't printed under "in file". You only get "In file: Framework" followed by nothing ;)

But besides that... this change is a HUGE improvement!

I've added it manually to my sources for now.

Thanks for the change!

veiko99:
@jicho I think it is a separate issue, you get "In file: Framework" when there is a syntax error somewhere.

jicho:
I know, but what I was trying to say is that in such a situation the template (view file) was visible of how it was rendered. Now you just see "In file: Framework" in stead of the file content.

veiko99:
Did it worked before? I couldn't reproduce. Could you show a ticket?

jicho:
Whoops, sorry... my mistake... :S Just compared Py2 and Py3, errors are the same.

I vote for a integration into the master version ;)

Issue #1987, opened 2018-08-15
<class 'KeyError'> "b'unique_key'" error in Python 3.5


jicho:
Hi,

When I'm logging in into Web2Py under Python 3.5 I get the error:
<class 'KeyError'> "b'unique_key'"

On this line:
```
if row and row[b'unique_key'] == to_bytes(unique_key):
```
in globals.py

Now... When I remove the "b" from b'unique_key' the error is gone, BUT on every login a new session record is created in the database, so I'm lossing session info on every redirect :(

I've fixed this problem by replacing the if statement with:
```
if row and row['unique_key'] == unique_key:
```

Since PY3 is doing a str comparison where, it's the same when using PY2...

No idea why the bytes compare is in there, but with session in a database combined with PY3 it's causing issues :)

jicho:
Seems to be fixed in https://github.com/web2py/web2py/pull/2047

PR #2047, opened 2018-10-29
'unique_key' shouldn't be saved as bytes

kvanzuijlen:
We started using a custom version of Web2Py for Python 3 compatibility. We want to start using the official release of Web2Py again but we were having issues that were fixed by changing this 2 lines. I think these 2 changes should be implemented in Web2Py but this should be thoroughly tested.

### DIFF
--- a/gluon/contrib/redis_session.py
+++ b/gluon/contrib/redis_session.py
@@ -186,8 +186,8 @@ class MockQuery(object):
             if rtn:
                 if self.unique_key:
                     # make sure the id and unique_key are correct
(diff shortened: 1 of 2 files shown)


jicho:
It looks like this covers also my issue: https://github.com/web2py/web2py/issues/1987

kvanzuijlen:
Hi @mdipierro,

Could you please release a new version? This will enable us to start using the latest bug fixes with a official build.

Thanks in advance,
kvanzuijlen

PR #2050, opened 2018-10-29
fixes #2049

### LINKED ISSUE #2049: KeyError on ExportClass.represented() call
**Describe the bug**
When you call _represented_ method from **ExportClass**, a ticket is created with below traceback:
<type 'exceptions.KeyError'> '<name of the first column in grid>'
```
  File "\web2py\gluon\sqlhtml.py", line 3683, in represented
    value = record[f]
  File "\web2py\gluon\packages\dal\pydal\objects.py", line 94, in __getitem__
    raise KeyError(key)
KeyError: '<name of the first column in grid>'
```

**To Reproduce**
1. Create class that inherits from **ExportClass**
2. Call self.represented() inside _export_ method in **your** class
3. See KeyError

**Expected behavior**
self.represented() returns ready-to-export array with resolved values of all columns (fields)

lkrn:


lkrn:
Unable to reproduce issue on 2.18.4 anymore.
Closing PR as not relevant.

PR #2054, opened 2018-11-04
Add aspect ratio to IS_IMAGE() validator

vinyldarkscratch:
There are a lot of cases where someone may want a perfect square, or a 16:9 screenshot, or whatever other aspect ratio you can think of.  Lately, my site's users have been unfortunately uploading profile pictures with aspect ratios that cause some display issues on certain parts of the website, and I want to enforce a perfect square on all profile pictures.

This PR adds an aspect ratio check to IS_IMAGE().  It defaults to `(-1, -1)`, which disables the check like `minsize`.


vinyldarkscratch:
Thanks for the pull!  💜

PR #2075, opened 2019-01-17
unicode() is not a standard python 3 function so using str() if pytho…

benlawraus:
unicode() is not a standard python 3 function so using str() if python 3 test is true
in 
gluon/languages.py

### DIFF
--- a/gluon/languages.py
+++ b/gluon/languages.py
@@ -326,6 +326,8 @@ def write_plural_dict(filename, contents):
 
 
 def sort_function(x):
+    if sys.version_info.major == 3: # python 3 compatibility
+        unicode = str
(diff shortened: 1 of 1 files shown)


ilvalle:
hi @benlawraus , would you mind to use the more standardized **to_something** functions available in gluon/._compact ?
In addition, this piece of code is not checked by any test, if you want to add one. see https://codecov.io/gh/web2py/web2py/src/master/gluon/languages.py#L307

vinyldarkscratch:
Looking into this more, it looks like the `to_unicode` function provides what you're looking for in less lines of code (and the rest of `languages.py` seems to be using it).  I'd recommend editing your commit to utilize this function.  :wink:

PR #2081, opened 2019-01-26
Only decode Markdown text when needed

vinyldarkscratch:
Python 3 treats strings as unicode, so there's no reason to decode them into unicode again.  This will only decode when a byte string is passed into the function.


ilvalle:
hi @vinyldarkscratch would you mind to use the more standardized to_something functions available in gluon/._compact ?

PR #2084, opened 2019-01-29
Fix login form clearing after failed login

### LINKED ISSUE #2083: Invalid Login - Redirect Not Setting "vars" to "request.post_vars"
**Issue**
In the Auth class login function, if an invalid login occurs (passes SQLFORM field validation), i.e. valid email format and password is appropriate length, then there is a redirect to the same login function to generate another SQLFORM but it passes in request.get_vars instead of request.post_vars. Since the invalid submission is due to a POST request, get_vars will be empty and the post_vars which the user submitted will be lost and thus return a new login form which is blank (email and password field are blank).

Solution:
 Change the redirect call to use vars=request.post_vars so the user submitted values are pushed through with the redirect.
Before Fix:
                        session.flash = self.messages.invalid_login
                        callback(onfail, None)
                        redirect(
                            self.url(args=request.args, vars=request.get_vars),
                            client_side=settings.client_side)

After Fix:
                        session.flash = self.messages.invalid_login
                        callback(onfail, None)
                        redirect(
                            self.url(args=request.args, vars=request.post_vars),
                            client_side=settings.client_side)

**To Reproduce**
1. Go to login page of welcome app
2. Submit a valid email format and a valid password length.
3. Results: email and password are currently reset to blank instead of remaining.

**Expected Results**
- Email and password should remain the same as when the user submitted them.

vinyldarkscratch:
Fixes #2083.

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -2637,7 +2637,8 @@ class Auth(AuthAPI):
                         # invalid login
                         session.flash = specific_error if self.settings.login_specify_error else self.messages.invalid_login
                         callback(onfail, None)
-                        request.post_vars['password'] = ""
+                        if 'password' in request.post_vars:
+                            del request.post_vars['password']
                         redirect(
                             self.url(args=request.args, get_vars=request.get_vars, post_vars=request.post_vars),
                             client_side=settings.client_side)
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -2637,8 +2637,9 @@ class Auth(AuthAPI):
                         # invalid login
                         session.flash = specific_error if self.settings.login_specify_error else self.messages.invalid_login
                         callback(onfail, None)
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Looks like Travis' failure was it choking on the scheduler for Python 3.6, unrelated to the changes of this PR.

leonelcamara:
Wouldn't this put a clear text password in the URL parameters creating a potential vulnerability?

vinyldarkscratch:
You have a good point @leonelcamara, wasn't thinking straight when I was writing this, whoops!  I've patched it up so that it will not remember the password in the `post_vars`, leaving the email/username.

mdipierro:
can you please change this by removing the password key if present, instead of setting it to ""

vinyldarkscratch:
Of course, @mdipierro!

PR #2107, opened 2019-02-26
Add new IS_FILE validator

### LINKED ISSUE #2073: IS_UPLOAD_FILENAME() takes regex for extensions -- why?
So I've been spending a lot of time with `IS_UPLOAD_FILENAME` and finding out that my regex has been problematic at least once every single time.  That's when I had a realization: "why?"  Why does the `extension` parameter require regex?  Sure, there may be those rare cases where someone needs to use regex to check against an extension, but most users would probably just want to insert a list of different extensions.  I feel that introducing regex by default implies more work for website developers.

I propose adding the ability to pass a list of extensions into `IS_UPLOAD_FILENAME` (essentially what #1875 has brought up), and I would like to know if others think the same.

### LINKED ISSUE #2106: IS_UPLOAD_FILENAME(extension="pdf") accepts "foo.pdfx" as valid filename
**Describe the bug**
As my concerns in #2073 expressed, I feel that the utilization of regex causes more issues than it solves them.  One that I have just discovered is that passing `extension="pdf"` into `IS_UPLOAD_FILENAME` lets the application accept `pdfx`, as a valid filename, as they both pass a `re.match()`.  However, the documentation explains this as if it must be an exact match.

**To Reproduce**
1. Create a form with an upload field.
2. Attach an `IS_UPLOAD_FILENAME(extension='png')` to the upload field.
3. Upload a file ending in `.png2000`.
4. Validator incorrectly accepts input.

**Expected behavior**
The validator should return an error when not an exact match.

**Proposed solution**
Unfortunately, this bug will most likely break backwards compatibility.  I propose creating a new validator, `IS_FILE` (or whatever we'd want to name it), to become the successor of `IS_UPLOAD_FILENAME`.  This new `IS_FILE` should accept a string, compiled regex, or list of strings/regex, rather than assuming that the input will always be a regex string.

vinyldarkscratch:
This fixes #2106.  More details within the issue.  Also fixes #2073.


vinyldarkscratch:
💜

PR #2145, opened 2019-03-12
don't use a regex to find exposed functions

### LINKED ISSUE #2142: Feature request: Add support to typing in function def return
web2py 2.18.3 returns and error when a function has typing in the def return.
Eg.
`def get_approval() -> Dict[str, SQLFORM]:`
or
`def get_approval() -> None:`

I suggest that support for it be added.

leonelcamara:
Fixes #2142


mdipierro:
Nice!

PR #2159, opened 2019-03-22
Change binary scripts to use PyInstaller with PY3 

nicozanf:
PyInstaller is currently the only one which is compatible with Python > 3.5.
See also https://github.com/nicozanf/web2py-pyinstaller for details


nicozanf:
Now it's compatible with PY2, also ;-)

PR #2167, opened 2019-04-03
Python Redis 3 compatibility

jpsteil:
Python redis recently moved from v2 to v3.  With this update it broke backwards compatibility.  See the details here:  https://pypi.org/project/redis/

V3 implements StrictRedis vs Redis which has been deprecated.

This change removes an unnecessary key from the parms passed to the pipe and changes the type of another.

### DIFF
--- a/examples/app.example.yaml
+++ b/examples/app.example.yaml
@@ -1,7 +1,7 @@
 #  For Google App Engine deployment, copy this file to app.yaml
 #  and edit as required
-#  See http://code.google.com/appengine/docs/python/config/appconfig.html
-#  and http://web2py.com/book/default/chapter/11?search=app.yaml
+#  See https://cloud.google.com/appengine/docs/standard/python/config/appref
+#  and http://www.web2py.com/book/default/chapter/13#Deploying-on-Google-App-Engine
 
 application: yourappname
 version: 1
--- a/gluon/contrib/redis_session.py
+++ b/gluon/contrib/redis_session.py
@@ -144,6 +144,8 @@ class MockTable(object):
             # add it to the index
             pipe.sadd(self.id_idx, key)
             # set a hash key with the Storage
+            kwargs.pop('locked', None)  # remove the 'locked' key from kwargs - not needed for redis
+            kwargs['modified_datetime'] = str(kwargs['modified_datetime'])
             pipe.hmset(key, kwargs)
             if self.session_expiry:
                 pipe.expire(key, self.session_expiry)
@@ -220,6 +222,8 @@ class MockQuery(object):
             if not self.db.r_server.exists(key):
                 return None
             with self.db.r_server.pipeline() as pipe:
+                kwargs.pop('locked', None)  # remove the 'locked' key from kwargs - not needed for redis
+                kwargs['modified_datetime'] = str(kwargs['modified_datetime'])
                 pipe.hmset(key, kwargs)
                 if self.session_expiry:
                     pipe.expire(key, self.session_expiry)
--- a/gluon/contrib/redis_session.py
+++ b/gluon/contrib/redis_session.py
@@ -145,7 +145,8 @@ class MockTable(object):
             pipe.sadd(self.id_idx, key)
             # set a hash key with the Storage
             kwargs.pop('locked', None)  # remove the 'locked' key from kwargs - not needed for redis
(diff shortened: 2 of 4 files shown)


niphlod:
this PR has not to do with v2 to v3 migration right ?
I get the "optimization" of not having the locked attribute stored (although it didn't hurt) but I don't get the stringification of modified_datetime

jpsteil:
I did that because I got an error about the type not being valid for
'modified_datetime' the same as I did for the 'locked' field.

Converting that datetime field to a string made the problem go away.

jpsteil:
Sorry, I might not have answer the entire question.  Correct, this is not
about Python 2 to 3.  It is about the python redis package going from v2 to
v3.

niphlod:
@jpsteil : and what about the other datetime then ?

jpsteil:
I received a type error on that field if I didn't convert to a string. I
will remove that and retest today.

Jim

jpsteil:
Simone

Here is the traceback without coercing modified_datetime to a string:

  File "/home/jim/dev/web2py/gluon/main.py", line 460, in wsgibase
    session._try_store_in_db(request, response)
  File "/home/jim/dev/web2py/gluon/globals.py", line 1236, in _try_store_in_db
    if not table._db(table.id == record_id).update(**dd):
  File "/home/jim/dev/web2py/gluon/contrib/redis_session.py", line
232, in update
    rtn = pipe.execute()[0]
  File "/home/jim/.local/lib/python2.7/site-packages/redis/client.py",
line 3520, in execute
    return execute(conn, stack, raise_on_error)
  File "/home/jim/.local/lib/python2.7/site-packages/redis/client.py",
line 3386, in _execute_transaction
    if EMPTY_RESPONSE not in options])
  File "/home/jim/.local/lib/python2.7/site-packages/redis/connection.py",
line 687, in pack_commands
    for chunk in self.pack_command(*cmd):
  File "/home/jim/.local/lib/python2.7/site-packages/redis/connection.py",
line 663, in pack_command
    for arg in imap(self.encoder.encode, args):
  File "/home/jim/.local/lib/python2.7/site-packages/redis/connection.py",
line 125, in encode
    "byte, string or number first." % typename)
DataError: Invalid input of type: 'datetime'. Convert to a byte,
string or number first.

mdipierro:
I am tempting to accept this, unless niphlod (nice to have you back) has any strong objection.

niphlod:
To be fair as long as it's consistent is good. It can't fail on just one
datetime??!

Il dom 7 apr 2019, 18:20 mdipierro <notifications@github.com> ha scritto:

jpsteil:
Anything I need to do to move this along?

-Jim

mdipierro:
I just would like some time to test it and why the datetime serialization
was not needed before.

mdipierro:
Niphlod is right. Why

kwargs['modified_datetime'] = str(kwargs['modified_datetime'])

but not

kwargs['crated_datetime'] = str(kwargs['created_datetime'])

or 

for key, value in kwargs.items():
     if not isinstance(value, (int, long, ...):
         kwargs[key] = str(value)

Can you point to some documentation of why this is necessary?

mdipierro:
I think we should adopt this solution https://stackoverflow.com/a/32322307 + json.dumps(default=str) + json.loads

jpsteil:
Sorry for not getting back to you sooner, I've been out sick all week.

Here is the relevant note in the python-redis v2 to python-redis v3 change notes:

https://github.com/andymccurdy/redis-py#encoding-of-user-input

> 
> redis-py 3.0 only accepts user data as bytes, strings or numbers (ints, longs and floats). Attempting to specify a key or a value as any other type will raise a DataError exception.
> 
> redis-py 2.X attempted to coerce any type of input into a string. While occasionally convenient, this caused all sorts of hidden errors when users passed boolean values (which were coerced to ‘True’ or ‘False’), a None value (which was coerced to ‘None’) or other values, such as user defined types.
> 
> All 2.X users should make sure that the keys and values they pass into redis-py are either bytes, strings or numbers.
> 

I don't have an answer for why this doesn't fail on created_datetime as well.  I'm going to do some more testing today after clearing my redis store.

Regarding the pickle solution, are you concerned about the overhead that will cause?

dlage:
The issues outlined here should already be fixed in master, including some testing that was added.

PR #2207, opened 2019-05-16
Migrator - add a command line option to ensure that all tables are migrated

dlage:
Several times using web2py I've come across errors during the migration process and I believe that isolating the process of migration will help ensure a new deploy has the correct database version waiting for it.

I wasn't sure where to put the code used to ensure that even the lazy tables are migrated so I left it isolated in a script file that is called when webpy.py is invoked with the --force_migrate option.
Any advice on how to create some tests to cover this option would be helpful.
```
python .\web2py.py -S welcome --force_migrate -D 10
```


leonelcamara:
This is useful to make an app migrate without having to change its config and using the command line.

yamandu:
Web2pyp migrations are always a headache when deploying in a collaborative project.
I liked this.

PR #2234, opened 2019-07-23
Fix encoding of pdflatex log file

jvanbraekel:
Hello,

Fix a crash in pdf exports ; when pdflatex is used and a log is generated. By default, open() use ASCII encoding codec, leading to crashs when non ASCII characters are present inside of the log file.  This inforce the use of UTF-8 encoding for log , allowing specials characters inside of the log.


jvanbraekel:
The second fix is for the failed login attempt. Previously, post vars :  username,password and signature where wrongly addded in the redirect link.  Now, only include the get_vars if any, thus don't send back wrong provided credentials.

jvanbraekel:
Third fix, enable greek symbols, by adding a package into the latex template. For ubuntu distribution, the textgreek.sty is installed with the textlive-science  APT package.

PR #2236, opened 2019-07-25
initialize env with abspath

dlage:
When running scripts, the path used is relative to the web2py root instead of the full path. This causes differences in migrations when using InDBMigrator - this causes pydal to try to repeat migrations already done.

### DIFF
--- a/gluon/shell.py
+++ b/gluon/shell.py
@@ -233,7 +233,7 @@ def run(
     errmsg = 'invalid application name: %s' % appname
     if not a:
         die(errmsg, error_preamble=False)
-    adir = os.path.join('applications', a)
+    adir = os.path.absdir(os.path.join('applications', a))
 
     if not os.path.exists(adir):
         if not cron_job and not scheduler_job and \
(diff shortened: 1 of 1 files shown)


mdipierro:
Can you explain the use case better? It seems to me it is the opposite. If the project is moved and you use the absolute path, it will trigger unwanted migrations.

dlage:
The issue is that a different path is used when running scripts in shell or when a web request is received. When using file storage it's really the same but when we use InDBMigrator each environment would follow a different pattern creating migration issues.

I agree that it would be better to use relative paths but since the (I believe mostly used way to migrate) is using the web-based option (since the --force_migrate option hasn't even been released) I thought this would cause less issues with current deploys (which is what worries me the most).

gluon/shell.py
```diff
@@ -234,7 +234,7 @@ def run(
     errmsg = 'invalid application name: %s' % appname
     if not a:
         die(errmsg, error_preamble=False)
-    adir = os.path.join('applications', a)
+    adir = os.path.absdir(os.path.join('applications', a))
```
mdipierro, line 237: OK. but should be os.path.abspath(....)

mdipierro:
Sorry for the very late reply.

mdipierro:
Sorry this took so long.

leonelcamara:
@mdipierro you forgot your own fix :) this should be abspath

mdipierro:
LOL

PR #2261, opened 2019-10-04
Update setup-web2py-fedora.sh

klausz:
# Centos 7.6 minimal, does not have zip installed. So there was ab error a few lines below. Unzip failed.
If possible also mention in the desctiption that this script is for Fedora and Centos. The filename can stay.


klausz:
There is an installation script for CentOS which works perfect. 
Therefore there is no need to change the fedora script.

PR #2280, opened 2019-12-11
Fix Issue 2274 - sessions using redis

### LINKED ISSUE #2274: sessions using redis broken
As per discussion on; https://groups.google.com/forum/?fromgroups#!topic/web2py/z1ee025ghRU

There are actually two issues.
1) adding () after the decode on line 185 of redis_session.py prevents a traceback occurring with messsage - AttributeError: 'builtin_function_or_method' object has no attribute 'decode'

2) web2py redis session code does not appear to retrieve the key successfully as it creates a new session for each page hit, ignoring the previous sessions created.  This makes the session in redis functionality unusable.  In other words, it creates the sessions (which can be seen in redis) but does not use them again and creates a new session per request.

python 3
python-redis library 2.x and 3.x
web2py 2.18.5 and master as at 31 oct 2019

Verified by another user as well (Leonel)

dlage:
Fix #2274 and add some basic tests of redis_session in gluon.tests.test_redis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
 before_script:
   - pip install coverage
   - pip install codecov
+  - pip install redis
 
 before_install:
   - mysql -e 'create database pydal;'
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ dist: "bionic"
 
 services:
   - mysql
+  - redis-server
 
 python:
   - '2.7'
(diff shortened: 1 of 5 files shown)


leonelcamara:
Nice!
@mdipierro ping

dlage:
As far as I could test this version of redis_session is compatible with either version 2 or 3 of pyredis. It should also deal correctly with old redis sessions so no additional actions are needed when upgrading.

PR #2282, opened 2020-01-15
Update dal to the current pydal master with major bugfix of JSON parser handoff

dlage:


mdipierro:
This should really be a pydal PR. I will fix it.

mdipierro:
This was merged in other PR

PR #2297, opened 2020-03-24
fix Py3 encoding problem

tyrbonit:
Fix for python 3.6
>>>str(b'123')
>>>"b'123'"
>>>str(b'123', encoding='utf-8')
>>>"123"


mbelletti:
@mdipierro 
In the patch, line 87 of serializer.py, **harattr** seems a typo instead **hasattr**

        return str(o) if harattr(str, 'decode') else str(o, encoding='utf-8')

PR #2306, opened 2020-04-10
Fix for undefined regex_tables

leonelcamara:


applications/admin/controllers/default.py
```diff
@@ -1261,7 +1261,7 @@ def plugin():
     defines = {}
     for m in models:
         data = safe_read(apath('%s/models/%s' % (app, m), r=request))
-        defines[m] = regex_tables.findall(data)
+        defines[m] = re.findall(REGEX_DEFINE_TABLE, data, re.MULTILINE)
```
mdipierro, line 1264: can we define regex_tables instead of compiling at every iteration? Thanks for looking into this.

PR #2372, opened 2021-01-09
Temporary fix for isAlive error

fcbarnard:
Temporary work around that takes care of the isAlive error. At least users can run web2py in debug mode


mdipierro:
No longer necessary but thanks.

fcbarnard:
I still get the same issue and have to always add this text. Not sure what is going on

PR #2397, opened 2021-08-30
Added install script for rhel8 based servers.

dmatic:
This script installs web2py with nginx on rhel8 based servers. During the installation the user can choose whether to use python2 or python3 for the setup.


mdipierro:
Thanks. Nice script.

PR #2419, opened 2022-01-11
Fix unique_key check w/ type bytes

### LINKED ISSUE #2330: Unable to login when using redis for storing sessions
Basically a new session is created on each page load so the login is never established.

Issue discussed at: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/web2py/6Ig5YVgvIsI/HpueAUELBgAJ

Confirmed with web2py versions from 2.18.5 up to 2.20.4.  Python versions 3.6 and 3.8

leonelcamara:
Fixes #2330


leonelcamara:
@mdipierro This is pretty crucial for web2py running with py3 to use redis

PR #2423, opened 2022-02-07
Fix travis fails to run tests

leonelcamara:
This fixes a bug in reply_to if you give it a string (e.g. you have only 1 email) instead of a list. This was causing an email test to fail.
This pull request also fixes the failing test_cache tests.

### DIFF
--- a/gluon/cache.py
+++ b/gluon/cache.py
@@ -49,6 +49,15 @@ except ImportError:
     HAVE_PSUTIL = False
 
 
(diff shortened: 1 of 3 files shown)


leonelcamara:
The only thing missing for web2py travis tests to be able to run is this fix in pydal.
https://github.com/web2py/pydal/pull/679

Then we need to update web2py's DAL again.

Afterwards the repos for both web2py and pydal need to be migrated from travis-ci.org to travis-ci.com otherwise the tests don't run and pull requests with breaking changes get merged.

PR #2430, opened 2022-02-16
Fixes bug for filenames with = in the url

leonelcamara:
See https://groups.google.com/g/web2py/c/emZGK0vLy_A/m/BOHVUXhxBAAJ


leonelcamara:
@mdipierro the bug was actually introduced here:

https://github.com/web2py/web2py/commit/4a3ccfdeab0670d42050a1b9709ec7355a13ada8

Can you review the fix.

Issue #2346, opened 2020-09-10
simplejsonrpc does not work with python3.8


iiijjjii:
In python3.8, urllib.request changed the way urllib.parse methods are imported                                                                                  
using leading underscores. As a result simplejsonrpc no longer works.
                                                                                                                                                                                                                                                        >>> from gluon.contrib.simplejsonrpc import ServerProxy                                                                                                         
>>> URL = "http://localhost:8000/app/webservice/call/jsonrpc"                                                                                                   >>> service = ServerProxy(URL, verbose=False)                                                                                                                   
AttributeError: module 'urllib.request' has no attribute 'splittype' 
                                                                                                                                                                                                                                                           
This can be fixed with this change in gluon/contrib/simplejsonrpc.py
-   import urllib.request as urllib                                                                                                                             
+   import urllib.parse as urllib

PR #2433, opened 2022-02-25
make simplejsonrpc compatible with python3.8+

timnyborg:
urllib's splittype and splithost were deprecated in 3.8

This also removes support for python2 in this contrib module (if required, it can be reintroduced, as urlparse was available as part of the urlparse module before integration into urllib.parse)

Fixes https://github.com/web2py/web2py/issues/2346

### DIFF
--- a/gluon/contrib/simplejsonrpc.py
+++ b/gluon/contrib/simplejsonrpc.py
@@ -18,12 +18,17 @@ __license__ = "LGPL 3.0"
 __version__ = "0.05"
 
 import sys
-from urllib import parse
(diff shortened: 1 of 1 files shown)


leonelcamara:
Can't you simply import urlparse from gluon._compat so it stays compatible? It changes almost nothing in your code and it would avoid the incompatibility.

timnyborg:
Excellent point.  I'll do so

timnyborg:
Updated with @leonelcamara's suggestion

leonelcamara:
@timnyborg thanks!

PR #2435, opened 2022-03-02
Update README.markdown

boninj:
Update dal path in the rm command example.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -27,7 +27,7 @@ If you have an existing repository, the commands below need to be executed at le
 
 If you have a folder gluon/dal you must remove it:
 
-    rm -r gluon/dal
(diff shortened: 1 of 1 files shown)


leonelcamara:
@boninj that path used to be a folder with the dal in web2py. When pyDal moved to its own repository and became a submodule, whoever still had the dal folder there needed to delete it or its name would clash with the dal.py file. You don't really need to delete gluon/packages/dal or gluon/dal anymore.

I actually think it may be fine to simply remove those lines from the readme as no one should be using a 7 year old version of web2py.

aahhh9 (approved):

mdipierro:
Fixed like Leonel suggested

PR #2445, opened 2022-08-29
Add DKIM support to emails

fambon:
add support to `Mail.send()` for the following spam-filters facilitating mail content:

- DKIM (requires install of `dkimpy`)
- List-Unsubscribe

example usage (global configuration e.g. via `models/db.py`):
```
dkim_key_path = myconf.get("dkim.key", default=default_dkim_key_path)
if path.exists(dkim_key_path):
    class dkim:
        key = open(dkim_key_path).read()
        selector = myconf.get("dkim.selector", default="s1024")

    mail.settings.dkim = dkim


list_unsubscribe = myconf.get("contact.email")
if list_unsubscribe:
    mail.settings.list_unsubscribe = list_unsubscribe
```

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -851,6 +851,38 @@ class Mail(object):
         return True
 
 
+def dkim_sign(payload, dkim_key, dkim_selector):
+
+    import dkim
+
+    # sign all existing mail headers except those specified in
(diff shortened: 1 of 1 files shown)


mdipierro:
Sorry it took me so long. Would be great if you could also make a PR to py4web about this. :-)

PR #2459, opened 2023-02-23
utils-compare fail safe

mcemeren:
compare function in utils.py assumes a and b are the same types, if not, it fails and throws an exception. Instead of throwing an exception, I recommend to fail safely and return False.


mcemeren:
Done.

PR #2496, opened 2024-10-18
 web2py.py

Balaadithya002:
added the code


mdipierro:
Thank you for your PR. I have to decline because of too many conflicts with the latest version but some of your suggested improvements have been incorporated already.

PR #2498, opened 2024-10-18
Update web2py.py

Manjusri-A:


mdipierro:
Declining this because code has changed a lot but some of your suggested improvements have been incorporated. thank you.

PR #2499, opened 2024-10-18
fix: ensure AppConfig reads configuration files with UTF-8 encoding

Pranika-Kalimuthu:


mdipierro:
There is no appconfig.py in the admin app. this PR does not make the case why we need it.

PR #2500, opened 2024-10-19
Successfully added  SECURITY.md file

Veeshavanachu:


jicho:
@mdipierro I'm not sure if this is a big issue, but... the SECURITY.md file isn't containing markdown, but good old HTML...

This should not be merged into the markdown file...

PR #2502, opened 2024-10-19
Create html5_form

2222k3060:
an HTML5 form in a single HTML file that collects basic user information


mdipierro:
This PR is has no explanation.

PR #2503, opened 2024-10-30
error with unicodeT in python3

KabluBR:
missing the code: from gluon._compat import unicodeT


mdipierro:
no longer necessary and there is no more _compat

PR #2515, opened 2024-12-18
Fix admin -> get_app info

nicozanf:
The ABOUT file of standard apps has changed its filename

### DIFF
--- a/applications/admin/controllers/default.py
+++ b/applications/admin/controllers/default.py
@@ -1087,8 +1087,12 @@ def about():
     """ Read about info """
     app = get_app()
     # ## check if file is not there
-    about = safe_read(apath('%s/ABOUT.web2py.txt' % app, r=request))
-    license = safe_read(apath('%s/LICENSE.web2py.txt' % app, r=request))
+    try:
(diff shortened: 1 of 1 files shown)


applications/admin/controllers/default.py
```diff
@@ -1087,8 +1087,12 @@ def about():
     """ Read about info """
     app = get_app()
     # ## check if file is not there
-    about = safe_read(apath('%s/ABOUT.web2py.txt' % app, r=request))
-    license = safe_read(apath('%s/LICENSE.web2py.txt' % app, r=request))
+    try:
+        about = safe_read(apath('%s/ABOUT.web2py.txt' % app, r=request))
+        license = safe_read(apath('%s/LICENSE.web2py.txt' % app, r=request))
+    except FileNotFoundError:
+        about = safe_read(apath('%s/ABOUT' % app, r=request))
```
mdipierro, line 1094: Do we need this fallback case? they do not exist any more in the repo.

nicozanf: I know, but I was thinking about old applications made by others

PR #2520, opened 2025-01-05
Remove cgi dependency

nicozanf:
It removes cgi dependency from web2py, which is needed for Python 3.13+ compatibility. Plus some additional docstrings.

**PLEASE NOTE:**  it apparently works fine, without any error, at least on Linux. But it's mainly made with ChatGPT and so it must be handled with caution. Check it carefully.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -5,14 +5,43 @@
 | Copyrighted by Massimo Di Pierro <mdipierro@cs.depaul.edu>
 | License: LGPLv3 (http://www.gnu.org/licenses/lgpl.html)
 
-Contains the classes for the global used variables:
+
+Web Request Processing Module for Web2py Framework
+
+This module handles HTTP request processing and variable management in the Web2py framework.
+It contains the classes for the global used variables:
 
 - Request
 - Response
 - Session
 
+It provides core functionality for:
+
+- HTTP request body parsing and caching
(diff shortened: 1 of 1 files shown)


leonelcamara:
This doesn't throw errors but in my tests it also doesn't seem to be working. I've managed to somewhat make it work with these changes:
```:diff
             # Handle multipart/form-data
             if content_type.startswith("multipart/form-data"):
-                # Extract boundary for parsing
-                boundary = content_type.split("boundary=")[-1].strip()
-                if not boundary:
-                    raise ValueError("No boundary in multipart form-data")
-
                 raw_data = body.read(content_length)
                 body.seek(0)

-                # Add boundary markers to match MIME format
-                raw_data = b"--" + boundary.encode("utf-8") + b"\r\n" + raw_data
+                # Boundary is embedded in content_type
+                raw_data = (
+                    "Content-Type: " + content_type + "\r\n" +
+                    "Content-Length: " + str(content_length) + "\r\n" +
+                    "\r\n"
+                ).encode("utf-8") + raw_data
+
                 parser = BytesParser()
-                msg = parser.parse(BytesIO(raw_data))
+                msg = parser.parsebytes(raw_data)
```

But I still don't seem to have a file being uploaded (I tested adding an extra upload field to auth_user and registering)

PR #2545, opened 2025-10-17
updated docker file for debian 

iamtalhaasghar:
have also added instructions on how to access admin panel for web2py instance running in docker


leonelcamara:
This looks good, but do we still need legacy cgi?

iamtalhaasghar:
Yeah, gunicorn depends on it i think or some other lib in web2py. @leonelcamara

PR #2549, opened 2026-02-26
fix: replace 201 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


haosenwang1018:
Scope is too broad for fast review; closing and can re-open as smaller chunks.

PR #2557, opened 2026-04-22
Fix host-header based trust bypass in appadmin local access control

metsw24-max:
appadmin currently determines whether a request is "local" by relying in part on the HTTP Host header and DNS resolution.

This is unsafe because the Host header is user-controlled and can be manipulated to bypass local-only access restrictions.

An attacker can send a request with a forged Host header that resolves to a loopback address causing the server to incorrectly treat the request as local and grant access to appadmin endpoints.

This patch removes Host-header and DNS-based trust from the access control logic and instead enforces locality strictly based on the actual client source address.

### Changes
- Removed Host header and DNS-based locality checks
- Introduced strict validation based on client source IP (loopback or configured trusted LAN prefix)
- Eliminated dependency on socket-based resolution for access control decisions
- Applied consistently across all appadmin controller variants (welcome, admin, examples)

### DIFF
--- a/applications/admin/controllers/appadmin.py
+++ b/applications/admin/controllers/appadmin.py
@@ -17,19 +17,12 @@ is_gae = request.env.web2py_runtime_gae or False
 global_env = copy.copy(globals())
 global_env['datetime'] = datetime
 
-remote_addr = request.env.remote_addr
-
-
-def is_loopback_address(ip_address):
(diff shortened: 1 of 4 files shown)


leonelcamara:
Why not just use request.is_local instead of is_loopback_address

metsw24-max:
> Why not just use request.is_local instead of is_loopback_address

I replaced the custom `is_loopback_address()` checks with `request.is_local` which uses the centralized local-request detection

PR #2560, opened 2026-04-23
Enforce path boundary validation in admin file operations

metsw24-max:
This PR strengthens path validation in the admin interface to prevent directory traversal and unauthorized file access outside the intended application boundaries.

### Changes
- Canonicalize paths using `os.path.realpath()` to resolve symlinks and traversal sequences
- Enforce strict directory boundary checks using `root + os.sep` to prevent prefix-based bypasses
- Apply validation consistently across:
  - `safe_open()`
  - `get_app()`
  - `listfiles()`
- Replace direct `open()` calls with `safe_open()` in admin flows (`report_progress`, `delete`)
- Add a regression test in `test_appadmin.py` to verify:
  - Legitimate paths are allowed
  - Path traversal (`../`) is blocked
  - Prefix bypass attempts are blocked
  - External/system paths are rejected

### DIFF
--- a/applications/admin/controllers/default.py
+++ b/applications/admin/controllers/default.py
@@ -89,13 +89,13 @@ def safe_open(a, b):
                 pass
         return tmp()
 
-    a_for_check = os.path.abspath(os.path.normpath(a))
+    a_for_check = os.path.realpath(os.path.normpath(a))
 
-    web2py_apps_root = os.path.abspath(up(request.folder))
+    web2py_apps_root = os.path.realpath(up(request.folder))
     web2py_deposit_root = os.path.join(up(web2py_apps_root), 'deposit')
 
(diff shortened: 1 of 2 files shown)


leonelcamara:
I'm not sure the change from abspath to realpath is correct, you could have these files in directories that are symlinks that resolve outside your application in production (e.g. an NFS folder shared by more than one webserver), I think we can assume that any symlinks inside the web2py directory are intentional.

I also find this code hard to read:

```python
if not any(a_for_check == root or a_for_check.startswith(root + os.sep) for root in allowed_roots):
```

I think the startswith is enough because you can't really `open` directories. You could basically keep that part of the code unchanged and simply add os.sep to web2py_apps_root and web2py_deposit_root.

PR #2563, opened 2026-04-24
encode non-ASCII characters in Subject header Fixes #2561

### LINKED ISSUE #2561: Mail.send()` does not encode non-ASCII characters in Subject header
### Description

`Mail.send()` fails with a `UnicodeEncodeError` when the `subject` parameter contains non-ASCII characters (e.g. German umlauts like `ü`, `ö`, `ä`).

The internal helper `encode_header()` is defined inside `Mail.send()` and correctly encodes non-ASCII strings as UTF-8 MIME headers using `email.header.Header`. However, it is never applied to the `subject` parameter — the subject is assigned directly to the payload without encoding:

```python
# gluon/tools.py, line 833
payload["Subject"] = subject  # non-ASCII characters not encoded
```

This causes the SMTP transport to raise:

```
Mail.send failure: 'ascii' codec can't encode character '\xfc' in position 172: ordinal not in range(128)
```

### Steps to reproduce

1. Configure `auth` with a translation file (e.g. `de.py`) that contains non-ASCII characters in a mail-related message, for example:
   ```python
   'Reset Password key': 'Passwort zurücksetzen'
   ```
2. Call `auth.request_reset_password()` or any `Auth` action that sends an email with this string as subject.
3. `mail.send()` returns `False`, `mail.error` contains the ASCII codec error.

### Expected behavior

`Mail.send()` should encode the subject using `encode_header()`, consistent with how other header fields are handled.

### Fix

Change line 833 in `gluon/tools.py` from:

```python
payload["Subject"] = subject
```

to:

```python
payload["Subject"] = encode_header(subject)
```

### Environment

- web2py version: 3.3.1
- Python version: 3.x
- OS: Linux

### Notes

This bug affects any application that uses non-ASCII characters in mail subjects, for example through translated `auth.messages` strings. It is particularly easy to trigger when using `T.force(language)` with language files containing umlauts or other non-ASCII characters.

trizer0:
Mail.send() fails with a UnicodeEncodeError when the subject parameter contains non-ASCII characters (e.g. German umlauts like ü, ö, ä).

Fixes #2561

### DIFF
--- a/gluon/tools.py
+++ b/gluon/tools.py
@@ -830,7 +830,7 @@ class Mail(object):
             to.extend(cc)
         if bcc:
             to.extend(bcc)
-        payload["Subject"] = subject
+        payload["Subject"] = encode_header(subject)
         payload["Date"] = email.utils.formatdate()
         for k, v in headers.items():
             payload[k] = v


leonelcamara:
@trizer0 thanks a lot for contributing! Your fix was not wrong, but it was incomplete as we needed to address the same problem with the other headers. Reviewing your PR pushed me into making a full fix. Feel free to contribute again.

PR #2567, opened 2026-04-25
Apply safe deserialization to globals and utils (follow-up to #2562)

metsw24-max:
Follow-up to #2562.

This PR extends the previously introduced safe deserialization approach to remaining unsafe pickle usages in gluon.globals `session handling` and gluon.utils `secure_loads` ensuring consistent protection against arbitrary code execution across all deserialization paths.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -71,6 +71,7 @@ import gluon.settings as settings
 from gluon import recfile
 from gluon.cache import CacheInRam
 from gluon.contenttype import contenttype
+from gluon.restricted import safe_load, safe_loads
 from gluon.contrib.multipart import MultipartParser, ParserError, parse_options_header
 from gluon.fileutils import up
 from gluon.html import PRE, TABLE, TR, URL, xmlescape
@@ -1117,7 +1118,7 @@ class Session(Storage):
                         )
                         portalocker.lock(response.session_file, portalocker.LOCK_EX)
                         response.session_locked = True
-                        self.update(pickle.load(response.session_file))
+                        self.update(safe_load(response.session_file))
                         response.session_file.seek(0)
                         oc = response.session_filename.split("/")[-1].split("-")[0]
                         if check_client and response.session_client != oc:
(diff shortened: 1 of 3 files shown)


leonelcamara:
I don't have time to review this properly today but I'm not sure about this one, I think it breaks compatibility for applications that save some objects in session.

I think at least it needs an escape hatch incompatible applications can use either by setting allowed classes or by making classes with a \_\_getstate__ and \_\_setstate__

What do you think @mdipierro ?

leonelcamara:
Yeah I don't think this is mergeable as it is., at the very least we would have to also allow datetime, decimal and pydal Row/Rows. If we're going to do this and break compatibility, we may as well only allow JSON and stop using pickle. 
  
If we really want to make sure the local sessions haven't been tampered with (which would imply a much larger security problem and access to your webservers) we should just use the same method we use with client side cookies and use secure_dumps/secure_loads.

I think this is going overboard for what is essentially zero benefit, you either already have access to the filesystem which is gameover, or you already have access to the DB which is also very close to gameover anyway.

The only real attack vector would be for people using redis sessions that are not autenticating the redis requests correctly or have some other exploitable vulnerability in their redis setup.

**My suggestion** would be to add `date`, `datetime`, `time`, `timedelta`, `uuid` and `decimal` to the allowed types, this would already make a lot of stuff work. Then add a safe_unpickle kwarg to session.connect (default False) this would allow people using setups that need more defense in depth (such as redis) to opt-in.
You would call doing something like this:
```python
session.connect(..., safe_unpickle=True, pickle_allowed_classes={"myapp.models": {"MyClass"}})
```

This would make the PR mergeable without breaking anyone and would allow people that need it extra protection. I would really like it.

metsw24-max:
Yeah I agree the current approach is a bit too restrictive and not worth breaking compatibility. Making this opt-in sounds much better.

I will update the PR to:
- allow common types
- add a safe_unpickle flag (default False)
- support pickle_allowed_classes for stricter setups

Will push changes shortly
Let me know if anyother changes to be done

leonelcamara:
@mdipierro I was waiting for changes to this PR this wasn't ready and broke compatibility

metsw24-max:
@leonelcamara i have applied the suggested changes should i open a follow up pr ?

leonelcamara:
@metsw24-max yes please

metsw24-max:
ok

metsw24-max:
@leonelcamara Opened a follow-up PR to address the compatibility concerns raised here: #2570

mdipierro:
Sorry Leonel for merging https://github.com/web2py/web2py/pull/2567.
I see https://github.com/web2py/web2py/pull/2570 is merged.

Should I release the latest master?

leonelcamara:
@mdipierro yes I think things are good now and we don't break backwards compatibility anymore.

PR #2569, opened 2026-04-28
Fix unsafe eval in appadmin

metsw24-max:
This PR fixes a security vulnerability in the admin interface (appadmin) where user-controlled input was passed to eval(), allowing arbitrary code execution.

## Fix

- Removed use of eval on user input
- Introduced a safe parsing function using `ast.literal_eval`
- Only literal values (strings, numbers, booleans, lists, dicts) are allowed
- Malicious expressions are not executed and are safely ignored

### DIFF
--- a/applications/admin/controllers/appadmin.py
+++ b/applications/admin/controllers/appadmin.py
@@ -109,7 +109,6 @@ def get_query(request):
 
 safe_eval_dict = gluon.utils.safe_eval_dict
 
-
 def query_by_table_type(tablename, db, request=request):
     keyed = hasattr(db[tablename], '_primarykey')
     if keyed:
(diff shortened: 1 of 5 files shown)


leonelcamara:
This has a of problem with commas

```python
safe_eval_dict('foo="hello world", bar="hello, world"')
{'foo': 'hello world', 'bar': '"hello'}
```

bar should be 'hello, world', maybe use ast.parse (you can pretend these are function keyword arguments) instead of split(',') you're already using the ast module anyway

It also needs to be addressed in all appadmins, maybe move safe_eval_dict to gluon.utils so it's easier to test and we can fix it only there next time

metsw24-max:
Moved `safe_eval_dict()` to gluon.utils and updated all `appadmin.py` controllers to use it. 
Added a regression test for comma-containing string values. `gluon.tests.test_appadmin` passes.

leonelcamara:
This is good!

PR #2573, opened 2026-04-28
Fix tickets

leonelcamara:
Two silent bugs broke the ticket system on Python 3.13+:

1. cgitb was removed from stdlib in Python 3.13, snapshot() raised ImportError (that was caught by a naked except) and silently stored {} instead.
2. XML objects produced by snapshot() were blocked by SafeUnpickler when loading tickets in the admin, causing those tickets to be silently skipped.


leonelcamara:
@mdipierro sorry we need a new release, the tickets were broken in ways that were not immediately apparent in simple tests if you weren't paying attentiion.

PR #2576, opened 2026-04-29
Add optional CSP nonce support for scripts and styles

jmestwa-coder:
### Summary

Adds **optional CSP nonce support** for inline `<script>` and `<style>` elements.

---

### What’s included

* `response.enable_csp()` to configure CSP headers
* `response.nonce` (per-response, lazy)
* Automatic nonce injection into:

  * `SCRIPT`
  * `STYLE`
  * inline resources via `include_files`
* Merges with existing CSP headers (no overwrite)
* Supports adding domains via parameters

---

### Design

* Implemented in `Response` (as discussed)
* Fully opt-in → no behavior change unless enabled
* Compatible with existing CSP headers
* Works with views, helpers, and `LOAD`

---

### Usage

```python
response.enable_csp(script_src=["https://cdn.example.com"])
```

---

### Backward compatibility

No changes unless `enable_csp()` is called.

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -568,6 +568,48 @@ class Response(Storage):
         self.delimiters = ("{{", "}}")
         self.formstyle = "table3cols"
         self.form_label_separator = ": "
+        self._csp_enabled = False
+
+    @property
+    def nonce(self):
+        if "nonce" not in self:
+            self["nonce"] = web2py_uuid()
+        return self["nonce"]
+
+    def enable_csp(self, nonce=True, **policies):
+        self._csp_enabled = True
+        existing = self.headers.get("Content-Security-Policy")
+        p = {}
+        if existing:
+            for directive in existing.split(";"):
+                directive = directive.strip()
+                if not directive:
+                    continue
+                bits = directive.split()
+                if bits:
+                    p[bits[0]] = bits[1:]
+
+        def merge(directive, sources):
+            if isinstance(sources, str):
+                sources = sources.split()
+            if directive not in p:
+                p[directive] = []
+            for s in sources:
+                if s not in p[directive]:
+                    p[directive].append(s)
+
+        if "default-src" not in p:
+            merge("default-src", ["'self'"])
+        if nonce:
+            n = self.nonce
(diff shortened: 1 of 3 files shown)


leonelcamara:
In **html.py**

```python
                n = current.response.nonce
                if n:
                    self["_nonce"] = n
```

I don't like code having guards for things that should never happen. response.nonce should never be empty with your implementation.

I also don't think we should be mutating (changing the attributes) of the SCRIPT/STYLE tags in the  xml method.

I would prefer something more like this.

```python
def _xml(self):
    fa, co = DIV._xml(self)
    if not self.attributes.get("_nonce"):
        from gluon.globals import current
        response = getattr(current, "response", None)
        if response and response._csp_enabled:
            fa += ' nonce="%s"' % response.nonce
    return fa, co
```

in **globals.py**

why are you using `getattr(self, "_csp_enabled", False)` instead of just self._csp_enabled.

Like I said I don't like unecessary guards.

```python
                    if (
                        getattr(self, "_csp_enabled", False)
                        and ("<script" in tmpl or "<style" in tmpl)
                    ):
```

You forgot the `link` tags which can have rel="stylesheet". We should probably use startswith as well.

I'm not loving the tmpl.replace approach anyway, I wonder if it's worth it to have a separate dict template_mapping_csp which we would use in these cases, but I don't think it's a blocker.

Finally don't forget the tests, we've been increasing coverage with your PRs, I like it.

jmestwa-coder:
Updated the patch based on your feedback:

* removed unnecessary guards
* avoided attribute mutation in `_xml`
* simplified `_csp_enabled` usage
* improved tag detection for include_files
* added tests

Let me know if anything else should be adjusted.

leonelcamara:
@jmestwa-coder I have added some failing tests, you may have to consider having a separate template_mapping dict in globals.py instead of the replace approach.

jmestwa-coder:
I’ve replaced the string replacement with a CSP-specific template mapping so nonce is injected correctly for script, style, and link tags. Tests for css, less, and js include_files now pass.

Let me know if you’d like any changes.

PR #2581, opened 2026-05-04
Add safe JSON encoder for script contexts to prevent XSS

metsw24-max:
This patch introduces a `jsjson` helper for safely embedding JSON inside <script> tags.

### Problem:
Embedding JSON using `XML(json.dumps(...))` allows attacker-controlled input to inject `</script>` sequences, breaking out of the script context and leading to XSS.

Example:
Input:
"</script><script>alert(1)</script>"

Before:
<script>var data="</script><script>alert(1)</script>";</script>

This results in script execution.

### Fix:
The new helper encodes "</" as "\u003c/" to prevent premature termination of the script element.

This follows established best practices used in modern frameworks.


leonelcamara:
I like this but you included test_output.txt by mistake

metsw24-max:
> I like this but you included test_output.txt by mistake

i by mistakely added that.....i added a new commit where i removed that file

leonelcamara:
@metsw24-max I'm sorry I should have reviewed it more carefully. I think this was already done if you used serializers.py json as it uses a [JSONEncoderForHTML](https://github.com/metsw24-max/web2py/blob/8b8c925ad929d0ccbbe768b593383dd23edd614b/gluon/serializers.py#L118)

Basically this is only an issue if you used the stdlib json instead of the one in serializers.json, could be worth it to have shorthand for it in html.py I guess.

metsw24-max:
> @metsw24-max I'm sorry I should have reviewed it more carefully. I think this was already done if you used serializers.py json as it uses a [JSONEncoderForHTML](https://github.com/metsw24-max/web2py/blob/8b8c925ad929d0ccbbe768b593383dd23edd614b/gluon/serializers.py#L118)
> 
> Basically this is only an issue if you used the stdlib json instead of the one in serializers.json, could be worth it to have shorthand for it in html.py I guess.

Thanks for the suggestion! I've added `json` to `html.py.__all__` with lazy loading to avoid circular imports. Now developers can use `from gluon.html import json` to get the HTML-safe version with `JSONEncoderForHTML` by default making it more convenient and discoverable.

leonelcamara:
That's not exactly what I meant, I don't want a json in html.py name clashing with serializers.json and stdlib json. I don't understand why you don't use serializers.json in jsjson, and I think it should probably be called SAFEJSON or something like that.

metsw24-max:
I have updated it so [gluon.html](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) no longer exports a [json](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) name. The safe helper is now [SAFEJSON](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and it uses [gluon.serializers.json](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) under the hood so it stays HTML-safe without clashing with stdlib [json](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) or [serializers.json](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

leonelcamara:
```python
jsjson = SAFEJSON
```
This alias is unecessary, jsjson never existed until now we don't need to be backwards compatible with it.

```python
    except Exception:
        # Fallback to string representation if object not serializable
        s = serializers_json(str(obj))
```

I really don't like this. Things that should fail, should fail loudly (No need for this try except).

```python
    # Prevent closing </script> or similar sequences from appearing raw
    s = s.replace("</", "\\u003c/")
```
This replace really isn't needed because serializers.json already substitutes these characters (see JSONEncoderForHTML).

Finally, the test should be in test_html.py and be called test_SAFEJSON.

With these changes I think SAFEJSON would be a nice shorthand from import json from serializers and using XML.

metsw24-max:
Done. I have made all the requested changes
[SAFEJSON](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is now a clean, simple shorthand from [serializers.json](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with [XML](vscode-file://vscode-app/c:/Users/akibs/AppData/Local/Programs/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) wrapping

leonelcamara:
Thanks!

PR #2582, opened 2026-05-04
Use safe_unpickle=True by default and allowlist Session for safe deserialization

netliomax25-code:
This change makes session deserialization use `safe_unpickle=True` by default.

Previously, session data was deserialized using `pickle` by default. Since pickle can execute arbitrary code during deserialization, this is unsafe if session storage is compromised (e.g., via database manipulation or file tampering).

This patch switches the default to safe unpickling and ensures compatibility by allowlisting `gluon.globals.Session`, which is required for normal session reconstruction.

Key points:
- Uses existing safe unpickling mechanism already present in the codebase
- Preserves backward compatibility by allowing `safe_unpickle=False` explicitly
- Limits allowed classes to only `Session`, avoiding broad exposure
- Minimal and localized change to session deserialization logic

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -1022,7 +1022,7 @@ class Session(Storage):
         cookie_key=None,
         cookie_expires=None,
         compression_level=None,
-        safe_unpickle=False,
+        safe_unpickle=True,
         pickle_allowed_classes=None,
     ):
         """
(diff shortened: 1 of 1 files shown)


leonelcamara:
This breaks old applications for very little security gain for most people. I'm all for defense in depth and have been accepting many merge requests that do just that, but I won't pay a high compatibility cost for a small defense in depth gain.  
  
People wanting to harden their applications can already do so with the new safe_unpickle option, and I recommend it in some cases, but it needs to be the app developer evaluating that.

PR #2584, opened 2026-05-07
Harden CSP policy parsing and directive token validation

jmestwa-coder:
## Summary

Harden `Response.enable_csp()` by validating CSP directive names and serialized directive tokens before merging or generating `Content-Security-Policy` headers.

## Changes

* Added validation for CSP directive names
* Added validation for serialized CSP directive tokens
* Reject malformed directive names and malformed serialized tokens before policy merge/generation
* Added handling for existing comma-separated CSP policy lists
* Reject non-string iterable directive tokens instead of silently coercing them

## Tests

Added regression coverage for:

* malformed directive-name rejection
* malformed serialized token rejection
* semicolon injection rejection
* comma-invalid token rejection
* existing CSP policy-list parsing
* non-string iterable token rejection
* valid CSP acceptance paths

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -127,6 +127,44 @@ template_mapping_csp = {
     "js:inline": '<script nonce="%s" type="text/javascript">\n%s\n</script>',
 }
 
+CSP_DIRECTIVE = re.compile(r"^[A-Za-z0-9-]+$")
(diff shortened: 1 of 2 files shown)


leonelcamara:
As an aside, I think it would be nice if the validation done by _validate_csp_directive was made optional of as I don't see much value in it for most apps (the keys are put there by the app developer supposedly)

jmestwa-coder:
Thanks, that makes sense. I can see the value is stronger on the serialized token validation side.

PR #2589, opened 2026-05-11
Remove unsafe eval from scheduler callable resolution

metsw24-max:
Replace `eval()`-based callable resolution in `scheduler.py` with explicit safe callable lookup.

The updated implementation:

* resolves dotted module paths using `importlib`
* resolves simple callable names from controlled namespaces
* validates resolved objects against allowed callable types before execution

## Testing

* verified callable resolution for dotted module paths
* verified valid scheduler task execution
* confirmed arbitrary expressions are no longer evaluated
* verified no syntax regressions in modified module

### DIFF
--- a/gluon/scheduler.py
+++ b/gluon/scheduler.py
@@ -529,7 +529,37 @@ def executor(retq, task, outq):
             result = dumps(_function(*args, **vars))
         else:
             # for testing purpose only
-            result = eval(task.function)(*loads(task.args), **loads(task.vars))
+            # Resolve the callable without using eval() to avoid executing
+            # arbitrary code. Support dotted module paths like
+            # 'package.module.func' or simple names present in globals,
+            # locals or builtins.
+            func_name = task.function
+            callable_obj = None
+            try:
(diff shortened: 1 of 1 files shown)


leonelcamara:
I don't see how this is safer than the eval.

I think we probably should just remove the else: branch entirely and fix any tests that needed it (are there any tests that need it?), the scheduler should always be executing with an app. else should just FAIL immediately

```python
else:
    raise ValueError(
        "task.application_name is required; cannot execute task '%s' without an app context"
        % task.function
    )
```

metsw24-max:
Thanks for the feedback, agreed.

I removed the no-app fallback path entirely in the scheduler executor. Tasks now require an app context, and if application_name is missing we fail immediately with a ValueError instead of trying to resolve and execute a callable outside app context.

I also reran the retry-failed scheduler test that was affected during iteration, and it passes with this change:
`gluon.tests.test_scheduler.TestsForSchedulerRunner.testRetryFailed`

So the behavior now matches your requirement: no execution without app context, fail fast.

PR #2596, opened 2026-05-19
selected-directory boundaries in admin file manager writes

jmestwa-coder:
## Summary

This change tightens admin file-manager path handling so file creation and upload operations remain constrained to the selected application directory.

## What Changed

- Added centralized path validation helpers:
  - `_check_app_path()`
  - `_join_app_path()`

- Validated that selected locations remain inside the authorized application root.

- Replaced direct path joins in `create_file()` and `upload_file()` with normalized boundary-checked joins.

- Preserved valid nested paths such as:
  - `views/default/index.html`
  - `static/js/app.js`

- Blocked traversal attempts that escape the selected directory or application boundary.

## Examples Now Rejected

- `../controllers/default.py`
- `../../admin/controllers/default.py`
- absolute path injection such as `/tmp/pwn.py`

## Tests

Added regression coverage that exercises the real admin controller helpers and verifies:

- valid nested paths continue to work
- traversal outside the selected directory is rejected
- cross-application escapes are rejected
- absolute-path injections are rejected

## Result

File-manager writes are now consistently normalized and constrained to the selected application directory instead of relying only on the broader `applications/` root boundary.

### DIFF
--- a/applications/admin/controllers/default.py
+++ b/applications/admin/controllers/default.py
@@ -15,6 +15,7 @@ import io
 import importlib
 
 from gluon.admin import *
+from gluon.admin import _check_app_path, _is_within_root, _join_app_path
 from gluon.fileutils import abspath, read_file, write_file
 from gluon.restricted import safe_load, safe_loads, TicketStorage
 from gluon.utils import web2py_uuid
@@ -77,10 +78,6 @@ def log_progress(app, mode='EDIT', filename=None, progress=0):
     if filename:
         safe_open(progress_file, 'a').write(
             '[%s] %s %s: %s\n' % (now, mode, filename, progress))
-
-
-def _is_within_root(path, root):
-    return path == root or path.startswith(root + os.sep)
 
 
 def safe_open(a, b):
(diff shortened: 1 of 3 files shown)


leonelcamara:
Can you check why the tests aren't running?

jmestwa-coder:
@leonelcamara , The earlier regression test approach was interacting too heavily with controller/runtime state and ended up leaking into multiprocessing-related imports during the full suite.

I’ve pushed a follow-up update that removes the runtime-heavy controller execution entirely and replaces it with a direct helper-based regression test using isolated request objects. The production boundary checks remain unchanged; this only simplifies and stabilizes the regression coverage.

niphlod:
maybe unrelated to this PR, but tests surface some warnings 

https://github.com/web2py/web2py/actions/runs/26220695753/job/77154606932?pr=2596#step:7:16

https://ci.appveyor.com/project/web2py/web2py/builds/54083856/job/32qsdmin0qg2qo1q#L62

leonelcamara:
Just a minor nuisance but why does _check_app_path _join_app_path have an underscore prefix, these aren't private functions, they're part of the admin modules "api" functions.  
  
@niphlod nice catch, we should fix it.

PR #2599, opened 2026-05-19
Fix XSS attribute breakout in sanitizer

metsw24-max:
## Summary

Fix a real XSS vulnerability in web2py’s sanitizer integration where attacker-controlled values in URL-bearing attributes (`href`, `src`, `background`) could break out of the attribute context and inject executable HTML attributes such as `onclick`.

The issue affected the documented:

- `gluon.sanitizer.sanitize`
- `XML(..., sanitize=True)`

APIs.

The root cause was inconsistent escaping behavior inside `yatl.sanitizer.XssCleaner.handle_starttag`: most attributes already used `xml.sax.saxutils.quoteattr`, but URL-bearing attributes were emitted using raw string interpolation.

Because `HTMLParser` decodes character references such as `&quot;` into literal `"`, attacker-controlled input like:

    <a href="https://example.com/&quot; onclick=&quot;alert(1)">x</a>

was sanitized into:

    <a href="https://example.com/" onclick="alert(1)">x</a>

creating a real executable event-handler attribute and enabling XSS.

After this patch, URL-bearing attributes are emitted using `quoteattr`, ensuring injected quotes remain safely contained inside the attribute value.

Example after fix:

    <a href='https://example.com/" onclick="alert(1)'>x</a>

`HTMLParser` correctly sees only a single `href` attribute.

### `gluon/sanitizer.py`

- Replace the previous one-line re-export with a documented compatibility wrapper.
- Import `XssCleaner`, `sanitize`, and `xmlescape` from `yatl.sanitizer`.
- Install a hardened replacement for `XssCleaner.handle_starttag`.
- Route `href`, `src`, and `background` through `xml.sax.saxutils.quoteattr`.
- Preserve existing sanitizer behavior for non-URL attributes and disallowed tags.
- Add detailed inline documentation describing the vulnerability and exploit mechanics.

### `gluon/html.py`

- Explicitly import `gluon.sanitizer` during initialization.
- Ensure the sanitizer hardening patch is installed whenever `gluon.html` loads.
- Remove reliance on indirect import ordering or optional yatl integration paths.

### `gluon/tests/test_html.py`

Add regression tests covering attribute-breakout payloads using entity-encoded quotes.

The tests:

- Pass malicious payloads through `XML(..., sanitize=True)`.
- Parse the resulting HTML using `html.parser.HTMLParser`.
- Assert that no injected event-handler attributes (`onclick`, `onerror`, etc.) appear on sanitized output.

### DIFF
--- a/gluon/html.py
+++ b/gluon/html.py
@@ -27,6 +27,9 @@ from urllib.parse import quote as urllib_quote
 from urllib.parse import urlencode
 
 from yatl import sanitizer
+# Importing gluon.sanitizer installs an XSS-hardening patch on
+# yatl.sanitizer.XssCleaner; do not remove.
+from gluon import sanitizer as _gluon_sanitizer  # noqa: F401
 from gluon import decoder
 from gluon.highlight import highlight
 from gluon.storage import Storage
--- a/gluon/sanitizer.py
+++ b/gluon/sanitizer.py
@@ -70,4 +70,21 @@ def _safe_handle_starttag(self, tag, attrs):
         self.open_tags.insert(0, tag)
 
 
-XssCleaner.handle_starttag = _safe_handle_starttag
+def _yatl_is_vulnerable():
+    """Return True if yatl's XssCleaner has the href attribute-breakout bug.
+
+    Tests with the known exploit payload so the patch is skipped automatically
+    once yatl fixes the bug upstream.
+    """
+    probe = XssCleaner(
+        permitted_tags=["a"],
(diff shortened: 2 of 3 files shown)


leonelcamara:
@metsw24-max ok but please make a pull request in web2py/yatl where the real bug lies
@mdipierro please check this one as it also affects py4web

metsw24-max:
@leonelcamara ok i will open in web2py/yatl

PR #2603, opened 2026-05-21
Prevent Host header poisoning in Auth-generated email links

metsw24-max:
This patch fixes a Host header injection issue affecting Auth-generated absolute URLs used in:

* password reset emails
* email verification flows
* invitation confirmation flows

Previously, when `auth.settings.host` was unset, `Auth.url(..., scheme=True)` could derive the host component from `request.env.http_host`, allowing attacker-controlled `Host` headers to influence security-sensitive links sent by email.

An attacker could trigger a password-reset request using a crafted `Host` header such as:

```text
Host: attacker.example
```

causing the victim to receive a reset link pointing to the attacker-controlled domain. When clicked, the reset token could be exposed to the attacker, enabling account takeover.

## Root Cause

`Auth.url(..., scheme=True)` delegated host resolution to `URL(...)` with `host=None`, which eventually fell back to:

```python
request.env.http_host
```

inside `gluon.rewrite.url_out`.

Although `Auth.__init__()` already validated hosts through `select_host(...)` when `host_names` was configured, the validated value was not reused during Auth email-link generation.

### `gluon/tools.py`

* Persist `host_names` into Auth settings
* Prevent `Auth.url(scheme=True)` from silently deriving hosts from the request `Host` header
* Use `select_host(...)` with `auth.settings.host_names` when resolving absolute URLs
* Raise HTTP 500 when absolute Auth URLs are requested without trusted host configuration

### `gluon/tests/test_tools.py`

Added regression coverage for:

* refusal without trusted host configuration
* explicit `settings.host` precedence
* allowlisted host acceptance
* allowlisted host mismatch rejection
* relative URL behavior remaining unchanged
* end-to-end password-reset link poisoning prevention

### DIFF
--- a/gluon/tests/test_tools.py
+++ b/gluon/tests/test_tools.py
@@ -1449,6 +1449,113 @@ class TestAuth(unittest.TestCase):
     # End Auth test
 
 
+class TestAuthHostHeaderPoisoning(unittest.TestCase):
+    """
+    Regression tests for Host-header poisoning in Auth.url(scheme=True).
+
+    Without this guard, an attacker can submit a password-reset request
+    with `Host: attacker.com` and the framework emails the victim a reset
+    link rooted at attacker.com -- the token leaks on click (CWE-640).
+    """
+
+    def _make_auth(self, http_host, host_names=None):
+        request = Request(env={})
+        request.application = "a"
+        request.controller = "c"
+        request.function = "f"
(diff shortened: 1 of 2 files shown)


leonelcamara:
@mdipierro please note that all applications need to have a host_names when they define Auth

PR #2604, opened 2026-05-21
Reject malformed multipart requests without caching partial form state

jmestwa-coder:
## Summary

Malformed `multipart/form-data` requests were previously treated as successful partial parses because `ParserError` was handled together with `StopIteration`.

This could leave `request.post_vars` in an inconsistent partially populated state when parsing failed after some fields had already been processed.

This change makes multipart parsing transactional:

- valid multipart requests are parsed and cached normally,
- malformed multipart requests now raise `HTTP(400)`,
- partial `post_vars` state is no longer cached on parser failure.

## Changes

- separated `StopIteration` and `ParserError` handling in multipart parsing
- raise `HTTP(400)` for malformed multipart bodies
- defer `_post_vars` assignment until parsing completes successfully
- added regression coverage for malformed multipart requests
- verified repeated access does not expose partial cached state

## Regression Test

Added a regression test that verifies:

- malformed multipart requests raise `HTTP(400)`
- `_post_vars` remains unset after parser failure
- repeated access continues to fail consistently
- no partial form state is exposed after failure

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -350,7 +350,7 @@ class Request(Storage):
         post_vars. application/json is also automatically parsed
         """
         env = self.env
-        post_vars = self._post_vars = Storage()
+        post_vars = Storage()
         body = self.body
 
         # if content-type is application/json, we must read the body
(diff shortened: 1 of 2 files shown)


leonelcamara:
This is as lenient as the old cgi.FieldStorage why the 400?

jmestwa-coder:
@leonelcamara , My thinking was that ParserError currently leaves request parsing in a failed state, but I agree the 400 changes the existing cgi.FieldStorage-style lenient behavior. I'll revisit whether the consistency issue can be addressed without changing those semantics.

PR #2606, opened 2026-05-22
Restrict admin file paths to selected app

XananasX7:
## Summary

  This tightens admin file operations so paths are scoped to the selected application root, not only the broader applications directory.

  Previously, admin create/upload/edit/delete paths were normalized only through the global applications root. In multi-user admin mode, a user authorized for one app could submit path traversal in a file
  location or filename and reach a sibling app under the same applications directory.

  ## Changes

  - Add reusable admin path join helpers that verify the resolved path stays under the intended base directory.
  - Scope admin file operations to the selected app root.
  - Preserve existing create/upload behavior for models, controllers, views, modules, static, private, and language paths.
  - Add regression coverage for sibling-app traversal.

  ## Validation

  - python3 -m py_compile gluon/admin.py applications/admin/controllers/default.py gluon/tests/test_compileapp.py
  - python3 -m unittest discover -s gluon/tests -p 'test_compileapp.py'
  - git diff --check HEAD~1..HEAD

### DIFF
--- a/applications/admin/controllers/default.py
+++ b/applications/admin/controllers/default.py
@@ -122,6 +122,31 @@ def safe_write(a, value, b='w'):
         safe_file.close()
 
 
+def safe_app_path(app, *members):
+    try:
+        return safe_join_admin_path(apath(app, r=request), *members)
+    except RuntimeError:
+        raise HTTP(403)
+
+
+def safe_app_location_path(app, location, *members):
+    try:
+        app_root = apath(app, r=request)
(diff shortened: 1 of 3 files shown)


leonelcamara:
We already have check_app_path, is_within_root and join_app_path, you need to consolidate, I don't want 2 sets of functions doing the same thing.

`location = os.path.relpath(location, app_root)`

in microsoft windows will throw a ValueError with different drives (e.g. C:\\ and D:\\), whereas you need the RuntimeError like in the other places.

PR #2608, opened 2026-05-23
Validate CSP nonce values before HTML and CSP serialization

jmestwa-coder:
## Summary

Adds defensive validation for `Response.nonce` before it is reused by built-in HTML and CSP serializers.

This prevents malformed nonce values from propagating into:

* CSP nonce attributes emitted by `SCRIPT`, `STYLE`, and `include_files`
* serialized CSP policy values generated by `enable_csp`

## Changes

* Added `_validate_csp_nonce()` helper for nonce format validation
* Added centralized nonce validation in `Response.nonce`
* Added regression coverage for malformed nonce values reaching:

  * `SCRIPT`
  * `STYLE`
  * `include_files`
  * `enable_csp`

## Notes

* The validation is intentionally narrow and preserves existing valid nonce formats generated by the framework
* The change focuses on preventing malformed nonce propagation and unsafe serialization paths
* No CSP policy semantics or rendering behavior were otherwise changed

### DIFF
--- a/gluon/globals.py
+++ b/gluon/globals.py
@@ -129,6 +129,7 @@ template_mapping_csp = {
 
 CSP_DIRECTIVE = re.compile(r"^[A-Za-z0-9-]+$")
 CSP_DIRECTIVE_TOKEN = re.compile(r"^[\x21-\x2B\x2D-\x3A\x3C-\x7E]+$")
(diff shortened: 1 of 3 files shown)


leonelcamara:
A developer using any kind of user input as a nonce has to be very confused and this would introduce a perfomance hit on every response.nonce read (even of automatic generated nonces). I don't see this threat as a real vulnerability.  
  
Please reopen if you disagree.

PR #2609, opened 2026-05-23
Harden temporary plugin filename handling in plugin_install()

jmestwa-coder:
## Summary

Harden `plugin_install()` temporary upload path handling by enforcing basename-only filenames before creating files in the `deposit` directory.

## What Changed

- Added `_safe_deposit_path()` to centralize temporary deposit filename validation.
- Reject filenames containing path components before temp-file creation.
- Preserved the existing `plugin_install()` failure contract (`False` on failure).
- Guarded cleanup logic to avoid secondary `unlink()` failures when file creation never occurred.

## Why

`plugin_install()` previously constructed the temporary deposit path directly from the provided filename before archive validation occurred.

Most upload flows already normalize filenames, but at least one URL-install path can pass source-derived filename data into `plugin_install()` without enforcing a basename invariant first.

This change moves the validation to the actual temp-file write boundary so invalid path-bearing filenames are rejected consistently before file creation.

## Tests

Added regression tests covering:

- traversal-like filenames return `False`
- empty / `.` / `..` names are rejected
- valid plugin basenames still succeed
- cleanup path remains safe when file creation fails
- invalid filenames never reach unpack logic

## Compatibility

- Existing valid plugin installs are unchanged.
- Failure behavior remains consistent (`False` on failure).
- No archive extraction behavior was modified.

### DIFF
--- a/gluon/admin.py
+++ b/gluon/admin.py
@@ -47,6 +47,14 @@ def _safe_extract_path(base_dir, member_name):
     return target_abspath
 
 
+def _safe_deposit_path(request, filename):
+    """Return a safe path for a temporary file in the deposit folder."""
+    basename = os.path.basename(filename)
+    if basename != filename or basename in ("", os.curdir, os.pardir):
+        raise RuntimeError("Invalid upload filename")
+    return apath("../deposit/%s" % basename, request)
+
+
 # TODO: move into add_path_first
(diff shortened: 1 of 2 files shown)


leonelcamara:
I assume safe_deposit_path can be used by anyone using the admin module, so it's not private and should not be prefixed by an underscore.

PR #2614, opened 2026-05-29
handling of HTTP suffix byte ranges

jmestwa-coder:
## Summary

Fix handling of HTTP suffix byte ranges in `stream_file_or_304_or_206()`.

## Problem

Suffix byte ranges such as:

```text
Range: bytes=-5
```

are currently interpreted as starting at byte `0`.

For a 10-byte file:

```text
Current:
Content-Range: bytes 0-5/10
Body: 012345

Expected:
Content-Range: bytes 5-9/10
Body: 56789
```

## Root Cause

When no start offset is present, the current logic falls back to `0`, causing suffix byte ranges (`bytes=-N`) to be treated as regular ranges from the beginning of the file.

## Fix

* Detect suffix byte ranges (`bytes=-N`) before applying the default start offset.
* Calculate the correct range for the last `N` bytes of the file.
* Preserve existing behavior for standard range requests.

### DIFF
--- a/gluon/streamer.py
+++ b/gluon/streamer.py
@@ -84,10 +84,14 @@ def stream_file_or_304_or_206(
             raise HTTP(304, **{"Content-Type": headers["Content-Type"]})
 
         elif request and request.env.http_range:
-            start_items = regex_start_range.findall(request.env.http_range)
-            if not start_items:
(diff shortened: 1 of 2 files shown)


leonelcamara:
This is a duplicate of #2613 which I like better

PR #2615, opened 2026-05-30
incorrect handling of HTTP suffix byte ranges

jmestwa-coder:
## Summary

Fix incorrect handling of HTTP suffix byte ranges in `stream_file_or_304_or_206()`.

A request such as:

```http
Range: bytes=-4
```

should return the last 4 bytes of the file. The current implementation interprets it as `bytes 0-4`, returning the first 5 bytes instead.

## Changes

* Add dedicated handling for suffix byte ranges (`bytes=-N`).
* Preserve existing range parsing behavior for all other range formats.
* Keep existing validation and error handling unchanged.


leonelcamara:
duplicate of #2613

PR #2623, opened 2026-06-04
Prevent XML element-name injection in serializers.xml()

metsw24-max:
Prevent XML element-name injection when serializing dictionaries with `serializers.xml()`.

Dictionary keys are converted into XML tag names through `TAG[key]`. Because tag names were derived directly from input values, specially crafted keys containing characters such as `<`, `>`, quotes, or whitespace could break document structure and inject arbitrary markup into the generated XML.

This change introduces `xml_safe_key()` and applies it to all key-to-tag conversions performed by `xml_rec()`.

## Changes

- Add `xml_safe_key()` helper.
- Preserve valid XML-safe element names unchanged.
- Normalize unsafe names before passing them to `TAG`.
- Apply sanitization to dictionary and list wrapper element names.

## Security Impact

Before:

```python
xml({
    "data": {
        'k]><script>alert(1)</script><x y=': "v"
    }
})
```

could inject attacker-controlled markup into the generated XML.

After:

- output remains well-formed XML
- breakout characters are neutralized
- arbitrary element injection is prevented

## Tests

Added regression tests covering:

- untrusted key sanitization
- preservation of valid keys
- `xml_safe_key()` behavior

All existing serializer-related test suites continue to pass.

### DIFF
--- a/gluon/serializers.py
+++ b/gluon/serializers.py
@@ -95,33 +95,29 @@ def custom_json(o):
         raise TypeError(repr(o) + " is not JSON serializable")
 
 
-# a tag name that is already a valid XML element name (a Name per XML 1.0,
-# restricted here to a conservative ASCII-plus-unicode-word subset) is emitted
-# verbatim; anything else has to be neutralised before it reaches a tag name
+# a valid XML element name (a Name per XML 1.0, restricted here to a
+# conservative ASCII-plus-unicode-word subset). A key carrying '<', '>', '"' or
+# whitespace would otherwise let (possibly attacker-controlled) data break out
+# of the element and inject arbitrary XML/markup.
 _XML_NAME_RE = re.compile(r"^[A-Za-z_:][\w.\-:]*$", re.UNICODE)
-_XML_NAME_BADCHAR_RE = re.compile(r"[^\w.\-]", re.UNICODE)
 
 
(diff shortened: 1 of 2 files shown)


gluon/serializers.py
```diff
@@ -95,13 +95,47 @@ def custom_json(o):
...
+    """
+    name = key if isinstance(key, str) else str(key)
+    if name == "":
+        # empty key is the "no wrapper element" sentinel used by xml_rec
+        return ""
+    if _XML_NAME_RE.match(name):
+        return name
+    name = _XML_NAME_BADCHAR_RE.sub("_", name)
+    # a valid XML name may not start with a digit, '.', '-' or be empty
+    if not name or not re.match(r"[A-Za-z_]", name):
```
mdipierro: Thanks for you PR. I do not believe we should transform this. I agree we should validate it.

metsw24-max: Good call — switched from transforming the key to validating it. xml_safe_key now raises ValueError for any key that isn't a valid XML element name (the empty-key "no wrapper" sentinel is preserved), so the serializer fails closed instead of silently rewriting/colliding keys. Tests updated accordingly.

metsw24-max:
@mdipierro any update on this?

PR #2627, opened 2026-06-08
fix exponential backtracking in multipart option-value regex

dxbjavid:
parse_options_header runs on the Content-Type header of every multipart POST before any authentication, and the quoted-string branch of its option-value regex lets a backslash be consumed by either the escape alternative or the non-quote alternative. an unterminated boundary value made of backslashes then backtracks exponentially, so a header of around 40 bytes can burn tens of seconds of CPU per request. limiting the non-quote alternative so a backslash is only matched by the escape rule keeps the two disjoint and makes matching linear, with no change to which values parse.


dxbjavid:
Thank you for the review and merge. Much appreciated

PR #2630, opened 2026-06-08
Escape username in ldap_auth uid-mode search filter (LDAP injection)

saddamr3e:
Fix LDAP filter injection in `gluon.contrib.login_methods.ldap_auth` when using `uid` mode with a service bind (`bind_dn` / `bind_pw`).

The LDAP search filter was previously constructed using unescaped user input:

```python
"(uid=%s)" % username
```

An attacker-controlled username containing LDAP filter metacharacters could alter the resulting query, enabling LDAP filter injection (CWE-90).

This change applies `ldap.filter.escape_filter_chars()` before interpolating the username into the search filter, matching the handling already used by the other filter-based authentication modes in the same file (`ad`, `company`, and `uid_r`).

## Security Impact

Before:

```python
username = "*)(uid=*"
filter = "(uid=*)(uid=*)"
```

After:

```python
filter = "(uid=\\2a\\29\\28uid=\\2a)"
```

The username is treated as data rather than LDAP filter syntax.

## Tests

Added a regression test in `gluon/tests/test_login_methods.py` that:

* Mocks the `ldap` module so the test runs without `python-ldap`
* Exercises the vulnerable `uid` authentication path
* Verifies that the filter passed to `search_s()` contains the escaped username
* Verifies that the injected LDAP clause is no longer present

### DIFF
--- a/gluon/contrib/login_methods/ldap_auth.py
+++ b/gluon/contrib/login_methods/ldap_auth.py
@@ -323,7 +323,10 @@ def ldap_auth(
                     con.simple_bind_s(ldap_binddn, ldap_bindpw)
                     dn = "uid=" + username + "," + ldap_basedn
                     dn = con.search_s(
-                        ldap_basedn, ldap.SCOPE_SUBTREE, "(uid=%s)" % username, [""]
+                        ldap_basedn,
+                        ldap.SCOPE_SUBTREE,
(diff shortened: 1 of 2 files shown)


saddamr3e:
@leonelcamara any update on this?

PR #2631, opened 2026-06-10
Fix XSS in Markmin media auto-link title handling

saddamr3e:
Fix a cross-site scripting (XSS) vulnerability in Markmin's media rendering path when a media title is recognized as an auto-URL and reused as the target of a wrapping `<a href>` element.

## Vulnerability

In `gluon/contrib/markmin/markmin2html.py`, `sub_media()` contains the following logic:

```python
if t and regex_auto.match(t):
    p_begin = p_begin + '<a href="%s">' % t
```

The title (`t`) is interpolated directly into an HTML attribute without escaping.

Two issues arise:

1. **Attribute breakout**

   * `regex_auto.match()` only validates a matching prefix.
   * Input such as:

     ```text
     http://x.com" onmouseover="alert(1)
     ```

     satisfies the guard because the URL prefix matches.
   * The full unescaped value is then inserted into `href`, allowing attribute injection and XSS.

2. **Unsafe URL schemes**

   * Unlike `sub_link()`, `sub_media()` does not reject `javascript:` URLs.
   * A title such as:

     ```text
     javascript://%0aalert(1)
     ```

     becomes a clickable JavaScript link.

### Example (before fix)

```text
[[http://x.com" onmouseover="alert(1) http://img.com/a.png img]]
```

Rendered as:

```html
<a href="http://x.com" onmouseover="alert(1)">
    <img src="http://img.com/a.png" />
</a>
```

This creates a live event handler and results in XSS.

## Fix

* Escape the auto-linked title using:

  ```python
  local_html_escape(t, quote=True)
  ```

  before inserting it into `href`.

* Reject unsafe URL schemes by applying:

  ```python
  not is_unsafe(t)
  ```

  consistent with the existing `sub_link()` behavior.

## Tests

Added `test_MARKMIN_media_autolink_title_xss()` covering:

* Attribute injection via unescaped quotes.
* `javascript:` URL rejection.
* Preservation of legitimate auto-URL image links.

### DIFF
--- a/gluon/contrib/markmin/markmin2html.py
+++ b/gluon/contrib/markmin/markmin2html.py
@@ -1436,8 +1436,14 @@ def render(
             p_end = "</p>" + pp
         elif p in ("left", "right"):
             style = ("float:%s" % p) + (";%s" % style if style else "")
-        if t and regex_auto.match(t):
-            p_begin = p_begin + '<a href="%s">' % t
+        if t and regex_auto.match(t) and not is_unsafe(t):
+            # t is an auto-url reused as the wrapping link target. regex_auto
+            # is matched with re.match (not anchored at the end), so a value
+            # like http://x" onmouseover="alert(1) still satisfies the guard;
+            # escape it (quote=True) so it cannot break out of href="..." and
+            # inject an event handler (XSS), and reject javascript: targets the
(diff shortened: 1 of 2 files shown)


saddamr3e:
@leonelcamara any update on this?

PR #2633, opened 2026-06-11
Prevent XML structure injection from dictionary keys in serializers.xml

jmestwa-coder:
## Summary

When serializing dictionaries to XML, `xml_rec()` emits dictionary keys directly as XML element names through `TAG[k]`.

While element values are escaped via `xmlescape()`, keys are inserted without any protection. Keys containing XML start-tag breakout characters (`<`, `>`, `/`, or whitespace) can alter the generated XML structure.

This change replaces only those breakout characters before emitting the element name, preventing keys from closing the current tag or introducing attributes.

## Changes

* Add `_xml_safe_tag()` helper for XML element names derived from dictionary keys.
* Replace XML start-tag breakout characters (`<`, `>`, `/`, whitespace) with `_`.
* Apply the helper when serializing dictionary keys in `xml_rec()`.
* Preserve all existing behavior for keys that do not contain breakout characters.
* Preserve the existing trailing `_` self-closing `TAG` convention.

## Examples

Before:

```xml
<x><script>alert(1)</script><y>data</x>
```

After:

```xml
<x__script_alert(1)__script__y>data</x__script_alert(1)__script__y>
```

## Tests

Added coverage for:

* Markup injection through dictionary keys.
* Attribute injection through dictionary keys.
* Nested dictionary key handling.
* Preservation of existing serializer behavior for normal keys.
* Preservation of the trailing `_` self-closing tag convention.

### DIFF
--- a/gluon/serializers.py
+++ b/gluon/serializers.py
@@ -95,11 +95,32 @@ def custom_json(o):
         raise TypeError(repr(o) + " is not JSON serializable")
 
 
+# A dict key is emitted as an XML element tag by xml_rec(). Replace only the
+# characters that let a key break out of its start tag -- '<', '>', '/' and
+# whitespace -- with '_', so an attacker-influenced key (e.g. "x><script>")
+# cannot close the tag or introduce attributes and inject markup. All other
+# characters are left untouched, which keeps existing output (including the
+# trailing-'_' self-closing TAG convention) unchanged for non-malicious keys.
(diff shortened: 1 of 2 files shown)


leonelcamara:
Thanks for this, @jmestwa-coder and good catch on the vulnerability. However, this overlaps with #2623, which targets the same issue (dict keys emitted as XML element names via TAG), so I'm going to close this in favor of that one as 2623 already addresses @mdipierro concerns of validating instead of transforming the xml.

PR #2634, opened 2026-06-11
Complete CSV formula neutralization in grid and service exports

jmestwa-coder:
## Summary

The recent CSV formula-injection hardening added `csv_safe()` protection to the `csv.writer()` paths in `Service.serve_csv()`, but Rows-backed exports still bypass that protection.

This affects:

* `Service.serve_csv()` when returning DAL `Rows`
* `ExporterCSV`
* `ExporterTSV`
* `ExporterCSV_hidden`
* `ExporterTSV_hidden`

All of these paths serialize through `Rows.export_to_csv_file()` / `as_csv()` and can emit spreadsheet formulas unchanged.

## Changes

* Add `csv_safe_rows()` to apply the existing `csv_safe()` logic to typed `Rows` values before CSV/TSV serialization.
* Apply it to Rows-backed exports in:

  * `Service.serve_csv()`
  * `ExporterCSV`
  * `ExporterTSV`
  * `ExporterCSV_hidden`
  * `ExporterTSV_hidden`
* Reuse the existing `csv_safe()` behavior instead of introducing a separate sanitization mechanism.

## Behavior

String values beginning with spreadsheet formula triggers are neutralized:

```text
=cmd|'/c calc'!A1
```

becomes:

```text
'=cmd|'/c calc'!A1
```

Numeric values retain their original type and output:

```text
-5
```

remains:

```text
-5
```

## Tests

Added coverage for:

* Rows-backed `Service.serve_csv()` exports.
* CSV exporter paths.
* TSV exporter paths.
* Type-aware behavior (`'=cmd` is neutralized while numeric `-5` is preserved).
* Idempotent application of `csv_safe_rows()`.

### DIFF
--- a/gluon/sqlhtml.py
+++ b/gluon/sqlhtml.py
@@ -4323,6 +4323,16 @@ class SQLTABLE(TABLE):
 form_factory = SQLFORM.factory  # for backward compatibility, deprecated
 
 
+def _csv_safe_rows(rows):
+    # Neutralize CSV/formula injection (CWE-1236) on the typed Row values
+    # before they are serialized, so string cells are defused while numeric
+    # columns keep their type. Lazy import avoids a gluon.tools <-> sqlhtml
+    # import cycle.
+    from gluon.tools import csv_safe_rows
+
+    return csv_safe_rows(rows)
+
+
 class ExportClass(object):
     label = None
     file_ext = None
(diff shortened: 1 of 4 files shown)


leonelcamara:
Duplicate of #2629 the other approach is slightly better

PR #2638, opened 2026-06-21
escape urls in autolinks before embedding in html

dxbjavid:
autolinks builds the embed markup for a matched url by dropping it straight into an href/src attribute (and into the web2py_component script string), and because the link regex is https?://\S+ a url that carries a double quote can close the attribute and inject an event handler, so user content rendered through expand_html ends up as reflected xss. the same unescaped interpolation is in image, audio, video, web2py_component and the plain-link branch of expand_one. this escapes the url with html.escape(quote=True) at each of those sinks, which lines up with the html.escape already used for the oembed query string just above.

### DIFF
--- a/gluon/contrib/autolinks.py
+++ b/gluon/contrib/autolinks.py
@@ -92,20 +92,20 @@ EMBED_MAPS = [
 
 
 def image(url):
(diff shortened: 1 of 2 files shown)


dxbjavid:
Thank you for reviewing and merging the PR. Appreciated.

PR #2640, opened 2026-07-01
percent-encode request.url in routes_onerror error redirects

dxbjavid:
When routes_onerror is configured, try_rewrite_on_error and try_redirect_on_error build the error handler redirect by dropping request.url straight into the request_url query parameter, even though the requested_uri value right next to it already goes through quote_plus. request.url is set from the url-decoded PATH_INFO, so a request path with a double quote breaks out of the href attribute in the "you are being redirected" body and an unescaped ampersand lets extra parameters be smuggled into the onerror handler URL. Encoding it with quote_plus at the three sites, matching requested_uri, keeps the value as data. I have added a regression test under test_routes that drives try_redirect_on_error with a quote-bearing path.

### DIFF
--- a/gluon/rewrite.py
+++ b/gluon/rewrite.py
@@ -281,7 +281,7 @@ def try_rewrite_on_error(http_response, request, environ, ticket=None):
                     status,
                     ticket,
                     quote_plus(request.env.request_uri),
-                    request.url,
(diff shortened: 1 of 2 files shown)


dxbjavid:
Thank you for reviewing and merging the PR. Appreciated.

PR #2641, opened 2026-07-02
reject reserved registration_key values in verify_email

dxbjavid:
verify_email looks up the account with table_user(registration_key=key) using the last URL segment as the key, but registration_key also holds the reserved states "", pending, disabled and blocked. An unauthenticated request to verify_email/disabled (or /blocked, /pending) therefore matches a disabled, blocked or approval-pending account and clears its key to "", quietly re-activating it. This refuses the empty and reserved key values before the lookup so only the random token issued at registration is accepted, and adds a regression test covering both the reserved values and a genuine token.


dxbjavid:
Thanks.

PR #2644, opened 2026-07-03
escape markmin meta attribute values in code, colour and cite

dxbjavid:
the optional id part of a code meta like ``text``:cls[id] is captured as [^]]* so it can contain a double quote, and that value is placed unescaped into three attributes of the code and meta rendering path: the <code> id, the colour <span> style, and the cite <a> href. a quote there closes the attribute and lets an attacker add an event handler, so any app that renders user-supplied markmin (comments, wiki pages, MARKMIN()) is exposed to stored xss, for instance ``hi``:cls[x" onmouseover="alert(1)] renders a live onmouseover handler. this escapes those values with quote=True, the same guard the link and media rendering already use, and adds a regression test alongside the existing markmin xss tests.


dxbjavid:
TY!

PR #2646, opened 2026-07-06
fix inverted can_read check in wiki search previews

dxbjavid:
Wiki.first_paragraph returns a page's first paragraph only when the current user cannot read it, and wiki search renders that paragraph for every page matching the query, so with manage_permissions enabled a searcher gets a preview of pages restricted to other groups. The same inverted check also means readable pages show no preview at all, so the teaser is effectively off in the default configuration. Swapping the guard to `if self.can_read(page)` shows previews for pages the user may read and withholds them otherwise.


dxbjavid:
TY!

PR #2655, opened 2026-07-13
enforce can_read on serialized wiki search results

dxbjavid:
**Restricted page bodies leaked on the serialised wiki search path**

With manage_permissions enabled the rendered search view withholds pages the searcher cannot read, but a request for any other extension (say _search.json) drops into a branch that returns as_dict() for every match, and as_dict() carries the full body, so a restricted page ends up disclosed to anyone able to search. The serialised branch now skips pages the requester cannot read, matching the check the rendered path already makes.


dxbjavid:
TY!

PR #2658, opened 2026-07-16
bound 206 responses when using the wsgi file wrapper

dxbjavid:
On a partial (206) response stream_file_or_304_or_206 sets Content-Length to the requested range length but, when web2py_use_wsgi_file_wrapper is enabled, hands the seeked file straight to the wsgi file wrapper, which streams it to the end of the file. The body then overruns the advertised length, so a Range request for the first hundred bytes of a large file returns the whole file behind a Content-Length of a hundred, and on a shared keep-alive connection the surplus bytes can bleed into the next response. Only use the wrapper for full-content responses and fall back to the bounded streamer for 206, which is what the non-wrapper path already does.

### DIFF
--- a/gluon/streamer.py
+++ b/gluon/streamer.py
@@ -133,7 +133,11 @@ def stream_file_or_304_or_206(
                 raise HTTP(404)
         headers["Content-Length"] = str(fsize)
         bytes = None
-    if request and request.env.web2py_use_wsgi_file_wrapper:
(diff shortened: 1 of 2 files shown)


dxbjavid:
TY!

leonelcamara:
@dxbjavid  We're the ones that need to thank :)

PR #2662, opened 2026-07-18
verify tls certificates in pysimplesoap transports

dxbjavid:
**TLS certificate verification is off by default in the SOAP transports**

All three transports in `gluon/contrib/pysimplesoap/transport.py` treat a caller-supplied `cacert` as the switch that turns verification on, so without one the chain and the hostname are both ignored and any https SOAP call is open to an active man-in-the-middle. The urllib2 transport is the worst of the three: it sets `CERT_NONE` unconditionally and raises on `cacert`, so there was no way to get a verified connection at all, and it is the transport selected on a stock install where neither httplib2 nor pycurl is present. My reading is that `cacert` was only ever meant to add a private CA rather than to enable verification, so I have made it behave that way and left the system trust store as the default.

I checked this against a local server presenting a self-signed certificate with a mismatched name: `urllib.request.urlopen` refuses it while the transport returned the body happily. The two tests in `test_contribs.py` cover both directions, since the second one would catch it if this turned out to be too strict for a private CA setup.

### DIFF
--- a/gluon/contrib/pysimplesoap/transport.py
+++ b/gluon/contrib/pysimplesoap/transport.py
@@ -125,7 +125,10 @@ else:
             if httplib2.__version__ >= "0.3.0":
                 kwargs["timeout"] = timeout
             if httplib2.__version__ >= "0.7.0":
-                kwargs["disable_ssl_certificate_validation"] = cacert is None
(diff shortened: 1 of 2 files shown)


dxbjavid:
Thanks.

PR #2677, opened 2026-08-17
reject blank passwords before ldap bind in ldap_auth and freeipa_auth

dxbjavid:
Both ldap_auth and freeipa_auth already refuse an empty password before binding, per RFC 4513 5.1.2, because an empty credential against a directory that permits an unauthenticated bind is accepted and the bind still returns success. The guard only compared the password against the empty string though, so a None password slipped past and reached the bind with an empty credential, logging the account in without a valid password; None is reachable when an app forwards a request variable that was never sent through login_bare to the registered login method. This widens the check to reject any falsy credential so it fires before a bind is attempted, which leaves a genuine password untouched.


dxbjavid:
Thanks!

PR #2687, opened 2026-08-30
scheduler: stop expanding an out-of-range cron range before rejecting it

tzh476:
`CronParser._rangetolist` takes the upper end of an explicit `min-max/step`
range straight from the cron expression, but `_sanitycheck` only runs after the
whole list has been built. A short expression therefore allocates a list
proportional to the number it names, and is then correctly rejected.

Measured with `tracemalloc` through `IS_CRONLINE`, the validator on
`scheduler_task.cronline`:

| expression | peak | time |
|---|---|---|
| `0-9999/1 * * * *` | 0.5 MB | 0.01 s |
| `0-999999/1 * * * *` | 48 MB | 0.6 s |
| `0-9999999/1 * * * *` | 480 MB | 6.9 s |
| `0-99999999/1 * * * *` | **4,800 MB** | **69 s** |

There is no cap on the number of digits, so the ceiling is not bounded by a
field width — each extra digit is another factor of ten.

To be clear about the reach, since it changes how much this is worth: the
`appadmin` route is admin-authenticated and `Scheduler.queue_task(cronline=...)`
is called by application code, so this is **not** an anonymous remote DoS. The
defect is that a validator spends gigabytes before rejecting a value, which is
the one thing a validator ought to do cheaply. Applications that expose
`cronline` to less privileged users get an amplifier on top.

### The fix, and why not a simple clamp

Generation now stops once a single out-of-range value is present. That one value
is all `_sanitycheck` needs in order to reject the expression, so every accepted
expression keeps its exact previous meaning.

Clamping the bound instead would change behaviour, which is worth spelling out
because it looks correct:

- `0-60/1` is **rejected** today — it yields `60`, which fails `_sanitycheck`.
- `0-60/7` is **accepted** today — step 7 yields `[0, 7, ..., 56]`, all in range.

So the bound alone cannot decide validity, and clamping it silently turns the
first case into a valid expression. Returning early on an out-of-range bound has
the mirror problem: it rejects the second case, which is legal.

I compared old and new over every field, every value in range plus five past it,
and every step — **1,430,136 expressions**. Accept/reject differs in **0** cases,
and the resulting lists differ in **0** cases.

### Test

`CronParserTest.test_hugerange` asserts the **peak allocation**, not just that
`ValueError` is raised. That distinction matters: the unpatched code raises the
same error with the same message, only after the allocation, so an `assertRaises`
alone passes either way. Without this change the new test fails with

```
AssertionError: 4799994202 not less than 1048576
```

It also pins `0-59/15` and `0-60/7` so the in-range behaviour, including a step
that runs past the end of the range, cannot regress.

### Test baseline

`python3 -m pytest gluon/tests/test_scheduler.py`

| | result |
|---|---|
| before (`git show HEAD:gluon/scheduler.py`) | 10 failed / 14 passed |
| after | 10 failed / **15** passed |

Those 10 failures are pre-existing and unrelated (`FileNotFoundError`; they need
a live DB and application directory) — I confirmed by running the same suite
against the unmodified file. `CronParserTest` passes 15/15.

### Not included on purpose

`gluon/newcron.py:202` has the same shape (`0-9999999/1` → 369 MB), but I left it
alone for two reasons: it has no `_sanitycheck` equivalent, so it currently
*accepts* out-of-range values and any bound there would change semantics rather
than harden; and its input is the local `crontab` file rather than a submitted
field value. Happy to follow up separately if you'd like it addressed.

---
Written with assistance from Claude (Anthropic); the measurements, the
1,430,136-expression comparison and the test baseline above were all run
locally and are reproducible from this branch.

### DIFF
--- a/gluon/scheduler.py
+++ b/gluon/scheduler.py
@@ -251,6 +251,7 @@ class CronParser(object):
                 s = s.replace("*", "1-12", 1)
             elif period == "dow":
                 s = s.replace("*", "0-6", 1)
+        ranges_max = dict(min=59, hr=23, mon=12, dom=31, dow=7)
         match = re.match(r"(\d+)-(\d+)/(\d+)", s)
         if match:
             max_ = int(match.group(2)) + 1
@@ -258,11 +259,25 @@ class CronParser(object):
         else:
             match = re.match(r"(\d+)/(\d+)", s)
             if match:
-                ranges_max = dict(min=59, hr=23, mon=12, dom=31, dow=7)
                 max_ = ranges_max[period] + 1
                 step_ = int(match.group(2))
         if match:
             min_ = int(match.group(1))
+            # The upper end of an explicit "min-max/step" range comes straight
+            # from the cron expression, so a short line can ask for a very long
+            # list: "0-99999999/1 * * * *" builds 100 million integers (~4 GB,
+            # ~40 s) before _sanitycheck rejects it for containing values above
+            # 59.  Stop generating once one out-of-range value is present: that
+            # single value is all _sanitycheck needs to reject the expression,
+            # and expressions that stay in range are unaffected.
+            limit = ranges_max[period] + 1
+            if max_ > limit:
+                if min_ >= limit:
+                    max_ = min(max_, min_ + 1)
+                else:
+                    # first multiple of step_ at or above the limit
+                    first_bad = min_ + -(-(limit - min_) // step_) * step_
+                    max_ = min(max_, first_bad + 1)
             retval = list(range(min_, max_, step_))
         else:
             retval = []
--- a/gluon/tests/test_scheduler.py
+++ b/gluon/tests/test_scheduler.py
@@ -8,6 +8,7 @@ import glob
 import os
 import shutil
 import sys
+import tracemalloc
 import unittest
 
 from gluon.dal import DAL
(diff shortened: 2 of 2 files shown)


tzh476:
Verification for this PR, since the only comment so far is the Codecov bot.

**The regression test is a real control — it fails without the fix.** Same test, same command, only
`gluon/scheduler.py` swapped between `master` and this branch:

```
with the fix:     1 passed  in  0.67s
without the fix:  1 failed  in 85.22s   (gluon/tests/test_scheduler.py:332 AssertionError)
```

The 85 seconds *is* the defect: `0-99999999/1 * * * *` makes `_rangetolist` build ~100M integers
before `_sanitycheck` rejects the expression for containing values above 59.

**No pre-existing test changes behaviour.** Full scheduler suite on both branches:

```
master (without patch):  10 failed, 14 passed
this branch:             10 failed, 15 passed   <- same 10, plus test_hugerange
```

The 10 failures are pre-existing on `master` and unrelated to this change (scheduler runner/API
tests). I'm reporting them rather than implying a clean suite.

**Legitimate expressions are untouched.** I enumerated every `lo-hi/step` combination across all
five periods (`min`, `hr`, `mon`, `dom`, `dow`) — 45,000+ expressions — and compared the original
and patched `_rangetolist` output:

```
legitimate expressions producing a different list: 0
```

That matters because the clamp bound scales with `ranges_max[period]`, not a fixed constant, so it
cannot truncate a valid range.

**End-to-end through the real class** (`CronParser(expr, datetime(...)).next()`):

| expression | result |
|---|---|
| `*/5 * * * *` | `2026-08-31 12:05:00` |
| `0-59/7 * * * *` | `2026-08-31 12:07:00` |
| `30 2 * * 1-5` | `2026-09-01 02:30:00` |
| `@daily` | `2026-09-01 00:00:00` |
| `0 0 1 1 *` | `2027-01-01 00:00:00` |
| `0-99999999/1 * * * *` | `ValueError` in 0.01 ms, peak RSS 47 MB |
| `70-99999999/1 * * * *` | `ValueError` in 0.01 ms |
| `* * * * 0-9999/1` | `ValueError` in 0.01 ms |

The rejection is unchanged — still `ValueError: Invalid cron value (...)`. The patch only stops the
list growing once one out-of-range value is present, which is all `_sanitycheck` needs to reject.

Happy to adjust anything about the approach or the comment wording.

tzh476:
One process note, not a ping about the review itself: **this PR's test workflow is queued but has never
been allowed to run.**

```
#2687 (this PR)   Tests  conclusion=action_required
#2686 (merged)    Tests  conclusion=success
```

Same single workflow, same repo — the only difference is that mine is a first-time-contributor PR, so
GitHub holds it behind **"Approve and run workflows"** until a maintainer clicks it. It is not a
failure and it will not clear on its own.

Flagging it because it changes what you are looking at: right now this PR shows you no test evidence,
which is the least convincing state a change can be in. One click and the repo's own `Tests` job will
either back the change or contradict me.

In the meantime, here is what I verified locally, which is the part that does not depend on CI. The
regression test is a real control — same test, same command, only `gluon/scheduler.py` swapped between
`master` and this branch:

```
with the fix:     1 passed  in  0.67s
without the fix:  1 failed  in 85.22s   (gluon/tests/test_scheduler.py:332 AssertionError)
```

The 85 seconds *is* the defect: `0-99999999/1 * * * *` makes `_rangetolist` build ~100M integers
before `_sanitycheck` rejects the expression for containing values above 59.

Nothing needed from me here — I will not follow up again on this.

